PR #131, opened 2015-01-09
JSON Response Files And Docs

nicky-johnstone:
Moving over the global JSON response implementation from the blog project to the Qcode-tcl library.

### DIFF
--- a/doc/global-json-response.md
+++ b/doc/global-json-response.md
@@ -47,15 +47,47 @@ Usage
 -----
 
 ### [qc::validate2model]
-This procedure validates input agains the data model and modifies the `record` portion of the response accordingly.
+This procedure validates input agains the data model and modifies the `record` object of the response accordingly.
 
 ### Custom Validation Handlers
-Custom validation handlers are used to manually validate input. As such, any manual validation should also set up the `record` portion of the response with the result of validation.
+Custom validation handlers are used to manually validate input. As such, any manual validation should also set up the `record` object of the response with the result of validation.
 
 ### Request Handlers
-Request handlers will often want to redirect the client to another URL after handling the request and therefore should set up the `action` portion of the response.
+Request handlers will often want to redirect the client to another URL after handling the request and therefore should set up the `action` object of the response.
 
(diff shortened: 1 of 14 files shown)


doc/global-json-response.md
```diff
@@ -0,0 +1,97 @@
...
+
+Details on setting up the response can be found in the [API].
+
+Usage
+-----
+
+### [qc::validate2model]
+This procedure validates input against the data model and modifies the `record` object of the response accordingly.
+
+### Custom Validation Handlers
```
nicky-johnstone: I actually started somewhat briefly on doing this but stopped to evaluate.

Currently, the registration processes are only for the blog project so adding the registration documentation here doesn't make sense at the moment. I considered moving the `register` and `validate` procs over to the library with this commit but then realised that there's also more stuff that needs moving over and thought it might perhaps be best done one pull request at a time?

Here's what I've established so far that still has to be moved over from blog to the library:

* new handler registration process
* `qc::handlers` API
* new filters
* new `conn_marshal` and the related request handlers
* `cast`, `castable`, and `is` ensembles as well as the `ensemble_handler`
* file management

Should I add the registration process and documentation into this pull request or in a later one?

doc/global-json-response.md
```diff
@@ -0,0 +1,97 @@
...
+            "message": ""
+        }
+    },
+    "message": {
+        "alert": {
+            "value": "Sorry, that email or password is not recognised."
+        }
+    },
+    "action": {
+
```
bvw: +1 for url rather than value. This might need to be a separate pull request in its own right.

doc/global-json-response.md
bvw: That's fine. Smaller chunks is better so I'll merge this one.

bvw:
I think we can move all the work you mentioned into the library.

We need to think about how we upgrade projects for the new cast ensemble and merge existing casts etc.

new handler registration process
qc::handlers API
new filters
new conn_marshal and the related request handlers
cast, castable, and is ensembles as well as the ensemble_handler
file management

PR #132, opened 2015-01-13
Safe HTML Error Report

nicky-johnstone:
* Utilities added for reporting unsafe HTML in some given text.
* Bugfix: `class` added as a safe attribute [but with restrictions in place] (https://github.com/qcode-software/qcode-tcl/blob/3c47be0423d3b1e2f6124464ba82875c0281f946/tcl/html.tcl#L445L450).
* `html2tcl` added from blog project.


tcl/html2tcl.tcl
```diff
@@ -0,0 +1,85 @@
...
+        return "<$tag_name [join $attributes]/>"
+    } elseif { [llength $attributes]==0 } {
+	return "<$tag_name>"
+    } else {
+	return "<$tag_name [join $attributes]>"
+    }
+}
+
+proc qc::h {tag_name args} {
+    #| Generate an html node
```
nicky-johnstone: I've noticed that `h` and the helper `h_tag` provide similar functionality to `qc::html` although not exactly the same implementation.

These have been moved over straight from the blog project so are in this file. Would they be best moved into `html.tcl`?

tcl/html.tcl
```diff
@@ -536,6 +538,114 @@ proc qc::safe_attributes_check {node} {
     return true
 }
 
+
+proc qc::html_error_report {text} {
```
bvw: Are these the best names for these procs ?

tcl/html2tcl.tcl
bvw: yes and I think we will drop qc::html and qc::html_tag from qcode-tcl-6 or at least deprecate.

tcl/html.tcl
nicky-johnstone: Was considering prefixing with `safe` to recognise association with a safe_html data type.

Some ideas: 

* safe_html_error_report
* safe_html_violations
* safe_html_errors
* safe_html_report

bvw: yes that sounds much better :+1:

nicky-johnstone:
Updated

tcl/html.tcl
```diff
@@ -37,11 +39,13 @@ proc qc::html2pdf { args } {
 }
 
 proc qc::html {tagName nodeValue args} {
+    #| Deprecated - use qc::h instead.
```
bvw: if there is a docs/proc this should also be marked as deprecated please.

tcl/html.tcl
```diff
@@ -60,6 +64,46 @@ proc qc::html_tag { tagName args } {
     }
 }
 
+proc qc::h_tag { tag_name args } {
```
bvw: is there any documentation for this proc ?

bvw: actually prob only need it for qc::h

nicky-johnstone: Yeah, I think that's the case.

These were pulled over from mla-www and I can't find any documentation. I'll add docs for `qc::h`.

nicky-johnstone:
Marked relevant docs as deprecated and added docs for `h` and `h_tag`

PR #139, opened 2015-01-14
Code migration from blog

nicky-johnstone:


tcl/register.tcl
```diff
@@ -0,0 +1,71 @@
+namespace eval qc {
+    namespace export register validate
+}
+
+proc register {method path proc_args proc_body} {
```
bvw: proc qc::register

nicky-johnstone:
Updated.

PR #143, opened 2015-01-16
Cast, Castable, Is additions and documentation.

nicky-johnstone:
* qc::cast decimal now accounts for precision and scale arguments to closer match PSQL data type.
* Various qc::is_x qc::cast_x and qc::is_x_castable moved to appropriate ensembles and originals marked as deprecated.
* Documentation for qc::is procs.

### DIFF
--- a/tcl/cast.tcl
+++ b/tcl/cast.tcl
@@ -586,15 +586,48 @@ namespace eval qc::cast {
         return $string
     }
 
-    proc decimal {string {precision ""}} {
-        #| Try to cast given string into a decimal value
+    proc decimal {args} {
+        #| Try to cast given string into a decimal value with the given precision and/or scale if present.
+        qc::args $args -precision ? -scale ? -- string
         set original $string
         set string [string map {, {} % {}} $string]
         if { [string is double -strict $string] } {
-            if { [string is integer -strict $precision] } {
-                return [qc::round $string $precision]
-            } else {
+            if { ! [info exists scale] && ! [info exists precision] } {
+                # Scale and precision not given.
                 return $string
+            } elseif { [info exists scale] && ! [info exists precision] } {
+                # Scale given but not precision.
+                if { [string is integer -strict $scale] && $scale >= 0} {
+                    return [qc::round $string $scale]
+                } else {
+                    return -code error -errorcode CAST "The scale must be a non-negative integer."
+                }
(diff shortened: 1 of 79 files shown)


daniel-qcode:
@nicky-johnstone 

Can you also audit tcl tests for places qc::is_* procs are used and change the code to use the new ensembles ?

nicky-johnstone:
Will do, yep.

nicky-johnstone:
The only places in tests taht any `qc::is` procs are used are in `is.test` and `cast.test` - the latter holding tests for `qc::is_period`.

Should the tests within `is.test` be updated to reflect the new ensemble usage or should they be left as they are and new tests added for the ensemble procs?

daniel-qcode:
@bvw 
Should the deprecated procs be changed to call the corresponding is ensemble ?

```
qc::is_uri_valid {uri} {
    return [qc::is uri $uri]
}
```

That way the code is centrailsed or should we keep deprecated procs with their own version of the code ?

bvw:
Yes we will move to the ensemble style syntax and deprecate the other style.

tcl/cast.tcl
```diff
@@ -688,4 +665,280 @@ namespace eval qc::cast {
             return -code error -errorcode CAST "Can't cast \"[qc::trunc $text 100]...\": not safe markdown."
         }
     }
+
+    proc values2model {args} {
```
bvw: I think this proc breaks the mould and should be 
qc::cast_values2model

nicky-johnstone:
On [another branch](https://github.com/qcode-software/qcode-tcl/compare/cast-castable-is-additions...update-is-calls) I'm updating any calls to `qc::is_x` to use the ensemble equivalent `qc::is x` including tests to be merged into this branch.

For the most part these are straightforward changes however there is a particular case where `checks`/`check` will probably fail: `varchar`.

The arguments for the old version: `qc::is_varchar string length` whereas the ensemble version has them switched around `qc::is varchar length string`. The reason being that the unknown subcommand handler for ensembles requires it to be this way due to how it passes back arguments to call the resolved subcommand.

I'm not sure of the changes required to `qc::check` to solve this problem.

Another change to `qc::check` that I'm unsure of is the [elseif clause on line 70](https://github.com/qcode-software/qcode-tcl/blob/e0d37b4c652e6b750e86b9215054d405d59209ce/tcl/check.tcl#L70)

PR #148, opened 2015-01-16
dont accept header values with ; but no parameters

davidqc:


bvw:
Could you please also submit a patch for branch qcode-tcl-5

PR #151, opened 2015-01-19
commit for discussion

daniel-qcode:


bvw:
Could we include this change in master too please.

PR #157, opened 2015-01-20
is_x -> qc::is x

nicky-johnstone:
* Updated occurrences of `is_x` calls to use the ensemble `qc::is x`.
* Updated `qc::is decimal` to account for precision and scale arguments.
* Updated and added tests.
* Tests all passed except crypt.test and db.test which didn't run due to missing package `pgtcl`.

### DIFF
--- a/tcl/is.tcl
+++ b/tcl/is.tcl
@@ -809,12 +809,12 @@ namespace eval qc::is {
         return [regexp -nocase {^[0-9a-f]*$} $string]
     }
 
-    proc url {string} {
+    proc url {args} {
         #| Checks if the given string is a URL.
         #| This is a more restrictive subset of all legal uri's defined by RFC 3986
         #| Relax as needed
-        args $args -relative -- url
-        default relative false
+        qc::args $args -relative -- url
+        qc::default relative false
         if { $relative } {
             return [regexp -expanded {
                 # path
--- a/tcl/is.tcl
+++ b/tcl/is.tcl
@@ -4,16 +4,12 @@ namespace eval qc {
 
 proc qc::is_boolean {bool} {
     #| Deprecated - see qc::is boolean
-    return [in {Y N YES NO TRUE FALSE T F 0 1} [upper $bool]]
+    return [qc::is boolean $bool]
 }
 
 proc qc::is_integer {int} {
     #| Deprecated - see qc::is integer
-    if { [string is integer -strict $int] && $int >= -2147483648 && $int <= 2147483647 } {
-	return 1
-    } else {
-	return 0
-    }
+    return [qc::is integer $int]
 }
 
 proc qc::is_pos {number} {
@@ -64,11 +60,7 @@ proc qc::is_pnz_int { int } {
 
 proc qc::is_decimal { number } {
     #| Deprecated - see qc::is decimal
-    if { [string is double -strict $number]} {
-	return 1
-    } else {
-	return 0
-    }
+    return [qc::is decimal $number]
 }
 
 proc qc::is_positive_decimal { number } {
@@ -102,7 +94,7 @@ proc qc::is_pnz_decimal { price } {
 proc qc::is_date { date } {
     #| Deprecated - see qc::is date
     # dates are expected to be in iso format 
-    return [regexp {^\d{4}-\d{2}-\d{2}$} $date]   
+    return [qc::is date $date] 
 }
 
 proc qc::is_timestamp_http { date } {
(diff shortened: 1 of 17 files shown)


bvw:
@daniel-qcode looks good to me.

test/is.test
```diff
@@ -2694,219 +2724,255 @@ bmZvIDEgMCBSCi9Sb290IDEzNiAwIFIKPj4Kc3RhcnR4cmVmCjExODgxMQolJUVPRgo=}
...
+        qc::is uri "http://test.com/$%^&m.pdf"
     } -result 0
 
-    test is_uri1.23 {is_uri invalid} -setup {} -body {
-        qc::is_uri_valid "///$%^&m.pdf"
+    test is-uri1.23 {is uri invalid} -setup {} -body {
+        qc::is uri "///$%^&m.pdf"
     } -result 0
 
+    test is-period-1.0 {qc::is period ""} {
```
daniel-qcode: Tests written like:
```
test some-test {some description} {test_code} {result}
```

or 

```
test some-test {some description} -body {test_code} -result {result}
```

Don't think it will work if we mix the two by using the -result flag but not the -body flag

tcl/is.tcl
```diff
@@ -534,10 +266,58 @@ namespace eval qc::is {
         return [expr {[string toupper $bool] in {Y N YES NO TRUE FALSE T F 0 1}}]
     }
 
-    proc decimal {number} {
+    proc decimal {args} {
         #| Checks if the given number is a decimal.
+        qc::args $args -precision ? -scale ? -- number
         if {[string is double -strict $number]} {
```
daniel-qcode: Is Postgresql's int range for decimals consistent with the range of doubles in TCL ?

tcl/is.tcl
```diff
@@ -534,10 +266,58 @@ namespace eval qc::is {
         return [expr {[string toupper $bool] in {Y N YES NO TRUE FALSE T F 0 1}}]
     }
 
-    proc decimal {number} {
```
daniel-qcode: @bvw Should this handle things like 1.5e10 ?

tcl/is.tcl
nicky-johnstone: PostgreSQL numeric/decimals:
> up to 131072 digits before the decimal point; up to 16383 digits after the decimal point

So they can get very large. Perhaps better to check the count of digits before and after a decimal point rather than checking if it's a double.

tcl/is.tcl
```diff
@@ -534,10 +266,58 @@ namespace eval qc::is {
...
+                    # Scale wasn't given so calculate it.
+                    set scale [expr {$precision - $left_count}]
+                }
+
+                if { $scale < 0 || ! [qc::is integer $scale]} {
+                    return -code error "Scale must be a non-negative integer."
+                }
+
+                set left_digit_max [expr {$precision - $scale}]
+                if { $left_count > $left_digit_max } {
```
daniel-qcode: This doesn't make any difference but I prefer something like:

```
if { $left_count > $left_digit_max } {
    return 0
} elseif { $right_count > $scale } {
    return 0
} else {
    return 1
}
```

or 

```
if { $left_count > $left_digit_max } {
    return 0
} 

if { $right_count > $scale } {
    return 0
} 

return 1
```

tcl/is.tcl
```diff
@@ -534,10 +266,58 @@ namespace eval qc::is {
         return [expr {[string toupper $bool] in {Y N YES NO TRUE FALSE T F 0 1}}]
     }
 
-    proc decimal {number} {
+    proc decimal {args} {
```
daniel-qcode: Should we raise an error if developer specifies scale but not precision ?

My first impression from tcl tests and examples was but precision and scale were both optional and not dependent on each other.

daniel-qcode: Sorry misread the code I'll take another look, think how it is just now might be fine

nicky-johnstone: Depends on how close we want to mimic the [PostgreSQL decimal/numeric type](http://www.postgresql.org/docs/9.4/static/datatype-numeric.html#DATATYPE-NUMERIC-DECIMAL) which only allows for precision and scale or just precision.

I'd guess that we do want to follow it as closely as possible?

tcl/is.tcl
bvw:
```
if { $left_count > $left_digit_max ||  $right_count > $scale } {
    return 0
} else {
    return 1
}
```

tcl/is.tcl
bvw: Anything that is valid in Postgresql should be valid here.
Our data types and literal representation should correspond as closely as possible with pg.

daniel-qcode: Something like 1.5e10 will be cast to 15000000000 by postgres.

So while the string "1.5e10" is not a valid decimal it is still castable so not sure if this proc should return o or 1 in this case ?

daniel-qcode: Currently "is decimal 1.5e10" will return 1.

But "is decimal 1.5e10 11 1" will cause an error

tcl/is.tcl
bvw: `expr {$x < pow(10,$scale-$precision)}`
to check if it fits in the space ?

tcl/is.tcl
daniel-qcode: If we wanted to handler then we could do something like this to convert them into strings:

```
if { [string first e $number] } {
    set number [qc::exp2string $number]
}
```

tcl/is.tcl
```diff
@@ -534,10 +266,58 @@ namespace eval qc::is {
...
         #| Checks if the given number is a decimal.
+        qc::args $args -precision ? -scale ? -- number
         if {[string is double -strict $number]} {
-            return 1
+            if { ! [info exists precision] && ! [info exists scale] } {
+                return 1
+            } elseif { ! [info exists precision] && [info exists scale] } {
+                set count [string length [lindex [split $number .] 1]]
```
daniel-qcode: Think we could use qc::intplaces here

tcl/is.tcl
```diff
@@ -534,10 +266,58 @@ namespace eval qc::is {
...
+                }
+                
+                set parts [split $number .]
+                set left_count [string length [lindex $parts 0]]
+                set right_count [string length [lindex $parts 1]]
+                if { $left_count == 0 } {
+                    # The leading zero wasn't given.
+                    set left_count 1
+                }
+
```
daniel-qcode:
```
set left_count [string length [cast_int $number]]
set right_count [qc::intplaces $number]
```

nicky-johnstone: Problem I can foresee here is that [PostgreSQL allows a massive 131072 digits before the decimal point](http://www.postgresql.org/docs/9.4/static/datatype-numeric.html#DATATYPE-NUMERIC-TABLE) so casting to integer would fail if it were out of integer range.

daniel-qcode: Actually that is not going to work because cast_int is range limited

daniel-qcode: yeah just realized that

On 20 January 2015 at 13:02, Nicky Johnstone <notifications@github.com>
wrote:

> In tcl/is.tcl
> <https://github.com/qcode-software/qcode-tcl/pull/157#discussion-diff-23220202>
> :
>
> > +                } else {
> > +                    return -code error "Scale must be a non-negative integer."
> > +                }
> > +            } elseif { [info exists precision] } {
> > +                if { $precision <= 0 || ! [qc::is integer $precision]} {
> > +                    return -code error "Precision must be a positive integer."
> > +                }
> > +
> > +                set parts [split $number .]
> > +                set left_count [string length [lindex $parts 0]]
> > +                set right_count [string length [lindex $parts 1]]
> > +                if { $left_count == 0 } {
> > +                    # The leading zero wasn't given.
> > +                    set left_count 1
> > +                }
> > +
>
> Problem I can foresee here is that PostgreSQL allows a massive 131072
> digits before the decimal point
> <http://www.postgresql.org/docs/9.4/static/datatype-numeric.html#DATATYPE-NUMERIC-TABLE>
> so casting to integer would fail if it were out of integer range.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/qcode-software/qcode-tcl/pull/157/files#r23220202>.
>



-- 
Kind Regards,
Daniel Clark
Qcode Software Limited
www.qcode.co.uk
Tel: +44(0) 1463 896 482

nicky-johnstone:
Updated with discussed changes.

Also modified behaviour to follow what PostgreSQL does when only precision is given (scale seems to be defaulted to 0):

```sql
blog[DEV]=# select 3.14::numeric(5);
 numeric 
---------
       3
(1 row)
```

```tcl
% qc::is decimal(5) 3.14
0
```

Is this appropriate or do we want the above to return true?

PR #158, opened 2015-01-20
cast_x -> qc::cast x

nicky-johnstone:
* Updated all `cast_x` to use ensemble `qc::cast x`
* Fixed bug with `qc::cast decimal`
* Exported `qc::cast epoch` as it's used by many other procedures.
* Updated, added, and run tests. All tests passed.


bvw:
ok with me.

PR #161, opened 2015-01-21
Update error.md

bvw:


nicky-johnstone:
As Daniel pointed out on the other pull request: `errorCode > error code` since we're trying not to rely on the global `errorCode`. It was something I had forgotten to change initially in this doc.

PR #163, opened 2015-01-21
Cast to domain base type where possible for qc::cast domain

nicky-johnstone:
`qc::cast domain` will now try to cast the given value to base type where possible.

### DIFF
--- a/tcl/cast.tcl
+++ b/tcl/cast.tcl
@@ -403,12 +403,10 @@ namespace eval qc::cast {
     proc domain {domain_name value} {
         #| Cast value to the domain $domain_name.
         set base_type [qc::db_domain_base_type $domain_name]
-        if { ![qc::is $base_type $value] } {
-            if { [qc::castable $base_type $value] } {
-                set value [qc::cast $base_type $value]
-            } else {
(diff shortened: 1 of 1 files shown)


tcl/cast.tcl
```diff
@@ -404,8 +404,14 @@ namespace eval qc::cast {
...
         if { ![qc::is $base_type $value] } {
-            return -code error -errorcode CAST "Can't cast \"[qc::trunc $value 100]...\": not a valid value for base type \"$base_type\" while checking \"$domain_name\" type."
-        } elseif { [qc::is $domain_name $value] } {
+            if { [qc::castable $base_type $value] } {
+                set value [qc::cast $base_type $value]
+            } else {
+                return -code error -errorcode CAST "Can't cast \"[qc::trunc $value 100]...\": not a valid value for base type \"$base_type\" while checking \"$domain_name\" type."
+            }
+        }
+
```
daniel-qcode: Do we need the outer if clause ?

nicky-johnstone: No. At least I can't think of a case where it's necessary. Removed and updated.

daniel-qcode:
@bvw 

Looks good to me.

This was my suggestion that casting to domain should always cast to the base type.

Eg 
Domain "poz_int" would have a base type of integer and a > 0 check constraint.
$ cast poz_int 1.4
1

Is this a sensible thing to do ?

before I think it would have raised an error

PR #164, opened 2015-01-23
Register paths fix

nicky-johnstone:
* `register` now allows for registering of paths i.e. `register GET /page`
* `qc::handlers` updated to choose a match with the least amount of colon variables
* `lappend` added to `qc::nsv_dict`

### DIFF
--- a/tcl/filter.tcl
+++ b/tcl/filter.tcl
@@ -11,7 +11,7 @@ proc qc::filter_validate {event {error_handler qc::error_handler}} {
     ::try {
         set url_path [qc::conn_path]
         set method [string toupper [qc::conn_method]]
-        # Check if this request is registered for this filter and a handler exists to validate against.
+        # Check if this request is registered and a handler exists to validate against.
         if { [qc::registered $method $url_path] && [qc::handlers exists $method $url_path] } {
             # Validate to data model
             qc::handlers validate2model $method $url_path
@@ -47,7 +47,7 @@ proc qc::filter_authenticate {event {error_handler qc::error_handler}} {
     set url_path [qc::conn_path]
     set method [string toupper [qc::conn_method]]
     ::try {
-        # Check if this request is registered for this filter
+        # Check if this request is registered.
         if { [qc::registered $method $url_path] } {
             if {[qc::cookie_exists session_id] && [qc::session_valid [qc::session_id]]} {
                 qc::session_update [qc::session_id]
--- a/tcl/filter.tcl
+++ b/tcl/filter.tcl
@@ -12,18 +12,21 @@ proc qc::filter_validate {event {error_handler qc::error_handler}} {
         set url_path [qc::conn_path]
         set method [string toupper [qc::conn_method]]
         # Check if this request is registered for this filter and a handler exists to validate against.
-        if { [qc::filter_handler_exists filter_validate $url_path $method] && [qc::handlers exists $url_path $method] } {
+        if { [qc::registered $method $url_path] && [qc::handlers exists $method $url_path] } {
             # Validate to data model
-            qc::handlers validate2model $url_path $method
+            qc::handlers validate2model $method $url_path
             # Custom validation
-            if {[qc::handlers validation exists $url_path $method]} {
-                qc::handlers validation call $url_path $method
+            if {[qc::handlers validation exists $method $url_path]} {
+                qc::handlers validation call $method $url_path
             }
             
             # Let the client know if there's a problem
             if {[qc::conn_open] && $method ne "GET" && ! [qc::record all_valid]} {
                 qc::return_result
                 return "filter_return"
+            } elseif { [qc::conn_open] && ! [qc::record all_valid] } {
+                # GET request failed validation
+                error "Couldn't validate the path \"$url_path\"" {} BAD_REQUEST                
             }
         }
 
@@ -45,7 +48,7 @@ proc qc::filter_authenticate {event {error_handler qc::error_handler}} {
     set method [string toupper [qc::conn_method]]
     ::try {
         # Check if this request is registered for this filter
-        if { [qc::filter_handler_exists filter_authenticate $url_path $method] } {
+        if { [qc::registered $method $url_path] } {
             if {[qc::cookie_exists session_id] && [qc::session_valid [qc::session_id]]} {
                 qc::session_update [qc::session_id]
             } elseif {$method ni [list "GET" "HEAD"] && [qc::cookie_exists session_id] && ! [qc::session_valid [qc::session_id]]} {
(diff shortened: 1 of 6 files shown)


tcl/register.tcl
```diff
@@ -68,4 +74,40 @@ proc qc::validate {method path proc_args proc_body} {
...
+    set method [string toupper $method]
+    if { [qc::nsv_dict exists registered $method] } {
+        return [qc::pattern_matches $url_path [qc::nsv_dict get registered $method]]
+    } else {
+        return false
+    }
+}
+
+proc qc::pattern_matches {url_path patterns} {
+    #| Checks if the given url_path matches any of the given patterns.
```
nicky-johnstone: There's likely a better name for this proc:

* `qc::path_matches_pattern`
* `qc::url_pattern_matches`
* `qc::url_patterns_match`
* `qc::matches_patterns`
* `qc::path_matches_patterns`
* `qc::path_patterns_check`
* `qc::check_path_patterns`

nicky-johnstone:
Checklist
------------------

- [ ] Does each proc have a #| comment to describe what it does ?
- [ ] Are there comments throughout the code ?
- [ ] Is each proc unit testable ?
- [ ] Are the proc names well chosen ?
- [ ] Are variable names well chosen ?

- [ ] List what testing has been done.
- [ ] UI changes - code reviewer should see changes.

nicky-johnstone:
Checklist complete.

tcl/handlers.tcl
```diff
@@ -3,14 +3,14 @@ namespace eval qc::handlers {
     namespace export call exists validate2model validation
     namespace ensemble create
 
-    proc call {path method} {
+    proc call {method path} {
```
bvw: What are the pros/cons of nested definition vs 
```tcl
namespace eval qc::handlers {
    proc call 
}
```
or
```tcl
proc qc::handlers::call
```

tcl/register.tcl
```diff
@@ -68,4 +75,40 @@ proc qc::validate {method path proc_args proc_body} {
     qc::nsv_dict set handlers VALIDATE $method $path args $args 
     qc::nsv_dict set handlers VALIDATE $method $path body $proc_body 
     qc::nsv_dict set handlers VALIDATE $method $path defaults $defaults
+}
+
+proc qc::registered {method url_path} {
+    #| Checks if the given method url_path is registered for the given filter.
```
bvw: `#| Checks if the given method url_path is registered.`

tcl/register.tcl
```diff
@@ -2,46 +2,53 @@ namespace eval qc {
...
+
+        # Update the handlers nsv array.
+        qc::nsv_dict set handlers $method $path proc_name $proc_name
+        qc::nsv_dict set handlers $method $path args $arg_names
+        qc::nsv_dict set handlers $method $path body $proc_body
+        qc::nsv_dict set handlers $method $path defaults $defaults  
+    }
 }
 
 proc qc::validate {method path proc_args proc_body} {
```
bvw: In `#|` comment terminology "extra validation" or "custom validation"
`#| Register a URL handler for extra validation.`
`#| Register a URL handler for custom validation.`

tcl/handlers.tcl
nicky-johnstone: ### Typical ensemble style
#### Pros
* Organisation/tidyness - This is the only reason that I can see as to why this style is usually used for ensembles.

#### Cons
* Currently, tags don't account for ensemble procs so they cannot be searched using `tags` commands in emacs.

tcl/filter.tcl
```diff
@@ -11,19 +11,22 @@ proc qc::filter_validate {event {error_handler qc::error_handler}} {
...
+            qc::handlers validate2model $method $url_path
             # Custom validation
-            if {[qc::handlers validation exists $url_path $method]} {
-                qc::handlers validation call $url_path $method
+            if {[qc::handlers validation exists $method $url_path]} {
+                qc::handlers validation call $method $url_path
             }
             
             # Let the client know if there's a problem
             if {[qc::conn_open] && $method ne "GET" && ! [qc::record all_valid]} {
```
daniel-qcode: How should we treat HEAD requests ?

tcl/handlers.tcl
```diff
@@ -55,29 +55,38 @@ namespace eval qc::handlers {
...
+                        # There is currently a colon variable command in the dictionary...
+                        if { $colon_variable_count < [dict get $command_dict variable colon_variable_count] } {
+                            # Prefer the command with the lower number of colon variables
+                            dict set command_dict variable [list command $command colon_variable_count $colon_variable_count]
+                        }
+                    } else {
+                        # No colon variable command currently stored.
+                        dict set command_dict variable [list command $command colon_variable_count $colon_variable_count]
+                    }                    
                 } else {
```
daniel-qcode: Do we need to test:

```
 [info exists command_dict] 
```

I think this variable is always guaranteed to exists because of the earlier set statement:

```
set command_dict {}
```

tcl/filter.tcl
nicky-johnstone: The idea is to only return the JSON response if the method is a type of `POST` request i.e. `DELETE` `RESTORE` `POST` etc.

`HEAD` should be treated like `GET` and excluded. Will update.

tcl/handlers.tcl
```diff
@@ -55,29 +55,38 @@ namespace eval qc::handlers {
...
                     # The pattern was an exact match
-                    dict set pattern_dict exact $command
+                    dict set command_dict exact [list command $command]
                     break
```
daniel-qcode: I can see that this was of setting the dict is consistent with earlier code when you were setting a dictionary value to be another dict.

Just an FYI that you can do this with just a dict set:

```
dict set command_dict exact command $command
```

tcl/handlers.tcl
```diff
@@ -174,22 +183,31 @@ namespace eval qc::handlers {
...
+                        # There is currently a colon variable command in the dictionary...
+                        if { $colon_variable_count < [dict get $command_dict variable colon_variable_count] } {
+                            # Prefer the command with the lower number of colon variables.
+                            dict set command_dict variable [list command $command colon_variable_count $colon_variable_count]
+                        }
+                    } else {
+                        # No colon variable command currently stored.
+                        dict set command_dict variable [list command $command colon_variable_count $colon_variable_count]
+                    }
                 } else {
```
daniel-qcode: Same here not sure that ```[info exists comnmand_dict]``` is neccessary

tcl/handlers.tcl
nicky-johnstone: Nope, we don't. Will update.

tcl/nsv_dict.tcl
```diff
@@ -82,4 +82,9 @@ namespace eval qc::nsv_dict {
             return [dict get $temp {*}[lrange $keys 1 end]]
         }        
     }
+
+    proc lappend {variable key values} {
+        #| Append the values to the key in an nsv_array
+        nsv_lappend $variable $key {*}$values
```
daniel-qcode: #| Append values to the list corresponding to the dictionary key stored in a nsv_array

tcl/register.tcl
```diff
@@ -2,46 +2,53 @@ namespace eval qc {
...
-    qc::nsv_dict set handlers $method $path args $args
-    qc::nsv_dict set handlers $method $path body $proc_body
-    qc::nsv_dict set handlers $method $path defaults $defaults    
-
-    if {$method eq "GET"} {
-        set http_methods [list HEAD GET]
+    set method [string toupper [lindex $args 0]]
+    set path [lindex $args 1]
+    # Add to registered nsv array
+    if { [qc::nsv_dict exists registered $method] } {
```
daniel-qcode: Haven't really looked into this but is there any reason we can't user qc::registered here instead of inspecting registered nsv array ourselves ?

tcl/handlers.tcl
bvw: Could qc::handlers::call be broken up to make components of it unit testable ?

tcl/register.tcl
nicky-johnstone: No reason. I think it was initially written this way because it was written before `qc::registered` was made. Updating.

tcl/handlers.tcl
nicky-johnstone: It's possible I think. Here's a breakdown of how each pattern is checked:

* Check that all colon variables in the pattern are present in the argument names.
* Check the number of parts from both pattern and path match and the number of arguments for the handler is not less than the number of colon variables in the pattern.
* Check each part of the pattern matches the corresponding path part and substitute colon variables in the pattern for their values from the path. 
* If the parts of the substituted path when put together match the pattern:
  * make the args
  * build the command
  * set the command_dict accordingly

I'll take a look and see how many dependent variables there are and how easy it might be to break it up.

nicky-johnstone:
Updated.

I'll create a new branch for the breakdown of `qc::handlers call`.

PR #181, opened 2015-02-02
qc::cast safe_html no longer casts to xml

nicky-johnstone:
This change is to fix a bug that makes safe HTML unsafe due to casting to XML. One particular instance was that casting to XML causes tags with no content between them to be void elements i.e. self-closing when they aren't supposed to be. This can cause problems of HTML breaking out of its container when used in particular cases.

For example:

```
% set html {<a href="test"></a>}
<a href="test"></a>
% qc::cast safe_html $html
<a href="test" />
```
The `a` tag is not a void element so shouldn't self-close as per w3 documentation: http://www.w3.org/TR/html-markup/syntax.html#syntax-elements

Checklist
------------

### Tcl
  - Testing done as changed in documentation plus a few other invalid and valid HTML snippets. All passed OK.

### DIFF
--- a/doc/procs/cast-safe_html.md
+++ b/doc/procs/cast-safe_html.md
@@ -17,16 +17,12 @@ Examples
 --------
 ```tcl
 
-% qc::cast safe_html "Hello World"
-<root>Hello World</root>
+% qc::cast safe_html {<a href="http://www.qcode.co.uk">Hello World</a>}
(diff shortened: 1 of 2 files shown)


bvw:
Would be good if we could enforce the safe_html domain in the database.
Can we tighten up the check of xml to validate in such a way as to prevent void elements like `<a/>`

PR #184, opened 2015-02-03
Introduce glob_recursive

PeterChaplin:


PeterChaplin:
Tcl
----
- [ ] Does each proc have a #| comment to describe what it does ?
- [ ] Are there comments throughout the code ?
- [ ] Is each proc unit testable ?
- [ ] Are the proc names well chosen ?
- [ ] Are variable names well chosen ?

- [ ] List what testing has been done.

daniel-qcode:
I'd prefer glob_recurive to recursively list matching files for each directory instead of creating a list of pattern for as specified max_depth.

daniel-qcode:
Would this approach work for directories that are symbolic links ?

PeterChaplin:
Is your concern based on the name of the proc, or would you still prefer a recursive search if the proc was named `glob_with_subdirectories`?

PeterChaplin:
I created a test folder "a", containing a file "testA.txt" and a sub folder "b". The sub-folder "b" contains a file "testB.txt" and a symbolic link named "c" which points back to "a".

With this approach, from inside "a", `qc::glob_recursive *.txt` returns a list of 11 items, the last of which was "b/c/b/c/b/c/b/c/b/c/testA.txt".

Honestly, I don't know if that's the most desirable behaviour or not, but it does show that this approach can follow symbolic links (and indicates the need to have some protection against looping, in this case a max_depth of 10 does that).

PeterChaplin:
I can think of three approaches to this proc:
1. What I currently have - build up patterns to a pre-specified max_depth, and call glob once with those patterns.
2. Iteration - create a list of directories, and use a `for` or `while` loop to iterate over the list from one end, while adding sub-directories to the other end.
3. Recursion - have the proc call itself for each sub-directory, and append the results of those calls to a glob of the local directory.

Building a pattern list necessitates a max_depth.
Recursion will also have a depth limit - tcl implements a recursion limit of 1000 (and even if we can increase that, we'd eventually hit a stack overflow error).
Iteration could run indefinitely, if we don't set an explicit limit to prevent that, though I suppose we'd eventually run out of memory to store the results in. It's probably best to specify a max_depth to prevent indefinite looping.

I don't know how the performance of the 3 would compare - I can't imagine any reason recursion would perform better (in fact I'm fairly certain it would be worse than iteration). I suspect that building up a single pattern and letting the OS handle traversal would perform best. That said, we may not need to worry about performance too much in this case.

daniel-qcode:
ok fair enough.

I think I would maybe prefer the name qc::find instead of qc::glob_recursive, but lets wit and get a third opinion from Bernhard when he is back.

tcl/util.tcl
```diff
@@ -542,3 +542,42 @@ proc qc::not_null {var} {
...
+
+proc qc::glob_recursive {args} {
+    #| Return names of files that match patterns, including files in sub-directories
+    # Usage: same as glob (see http://www.tcl.tk/man/tcl8.5/TclCmd/glob.htm),
+    # with additional max_depth switch
+    qc::args $args -join -nocomplain -tails -path ? -directory ? -types ? -max_depth 10 -- args
+
+    # Use a single glob call, but extend each pattern passed in with */ prefixes
+    # (eg. *.html becomes *.html */*.html */*/*.html etc.)
+    if { [info exists join] } {
```
bvw: Do we need a "-join" option ? 
Can responsibility for this feature be pushed up to the caller ?

`qc::glob_recursive some folder foo`
`qc::glob_recursive some/folder/foo`

Does it make sense for this ?
`qc::glob_recursive path1/folder/ path2/folder`

tcl/util.tcl
```diff
@@ -542,3 +542,42 @@ proc qc::not_null {var} {
...
+    }
+    set patterns $base_patterns
+    foreach depth [.. 1 $max_depth] {
+        foreach pattern $base_patterns {
+            lappend patterns [join [list {*}[lrepeat $depth *] $pattern] /]
+        }        
+    }
+
+    # Options to be passed on to glob
+    set glob_options [list]
```
bvw: Terminology `switches` `options`

`set switches [list]`
`set options [list]`

....
`glob {*}switches {*}$options -- {*}$patterns`

tcl/util.tcl
PeterChaplin: I included -join because `glob` has it, and I figured anyone using `glob_recursive` would expect it to have the same options as `glob` unless there was a good reason not to.

That said, I can't figure out why the developers of tcl felt the need to include this option in the first place, so I guess I'd be equally happy to take it out altogether.

(Incidentally, `glob -join path1/folder/ path2/folder` is equivalent to `glob path1/folder//path2/folder`. I don't know if that makes sense, but it is how tcl does it).

tcl/util.tcl
PeterChaplin: Fixed

tcl/util.tcl
bvw: Fair enough. Original Tcl is puzzling :frowning:

bvw:
Could we also have this in qcode-6 please.

PR #185, opened 2015-02-05
xlsx support

PeterChaplin:
Requires:
```
sudo apt-get install libexcel-writer-xlsx-perl
```

### DIFF
--- a/tcl/excel.tcl
+++ b/tcl/excel.tcl
@@ -4,7 +4,7 @@ namespace eval qc {
 
 proc qc::excel_file_create {args} {
     #| Creates an xls or xlsx file using the information provided
-    # type: "xls" or "xlsx", defaults to "xls"
+    # type: "xls" or "xlsx", defaults to "xlsx"
     # data: a list of lists containing a grid of cell values
     # formats: a dict of class definitions
     # class definitions: a dict of name-value pairs
--- a/tcl/excel.tcl
+++ b/tcl/excel.tcl
@@ -18,7 +18,7 @@ proc qc::excel_file_create {args} {
     default row_meta {}
     default cell_meta {}
     default timeout 1000
-    default type "xls"
+    default type "xlsx"
     switch $type {
         "xls" {
             set writer "Spreadsheet::WriteExcel"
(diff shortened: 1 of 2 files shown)


tcl/excel.tcl
```diff
@@ -10,13 +10,25 @@ proc qc::excel_file_create {args} {
     # column_meta: nested dict, eg. {1 {class "foo" width 20} 5 {width 50}}
```
daniel-qcode: Can you update comment to ```Creates an xls or xslx file ...``` or something along those lines

tcl/excel.tcl
```diff
@@ -10,13 +10,25 @@ proc qc::excel_file_create {args} {
...
+        "xls" {
+            set writer "Spreadsheet::WriteExcel"
+        }
+        "xlsx" {
+            set writer "Excel::Writer::XLSX"
+        }
+        default {
+            error "Unsupported excel type"
+        }
+    }
```
daniel-qcode: We should maybe raise an issue to upgrade all projects to use the new xlsx format then we could make this the default.

daniel-qcode:
Looks good just a few comments to address.

Then we will also need a similar pull request for qcode-tcl-6 (master)

daniel-qcode:
Could you also update ```qc::mime_file_extension``` and ```qc::mme_type_guess``` for .xlsx  files ?

PeterChaplin:
Updated mime types and comments

tcl/excel.tcl
bvw: Can we make xlsx the default `$type` as part of this pr ?
The xls format may become deprecated in time too.

PeterChaplin: This proc is also used by `/new_crawley_bin_nos.xls` for tlc-erp.
If we're not worried about that, I'm happy to make this change.

bvw: Good catch ! 
Could we change the tlc-erp code to specify `$type` and rollout the new qcode-5 package with that ?
Will then also need to go into qcode-6.

PeterChaplin:
I've made xlsx the default type.

PR #190, opened 2015-02-10
refactored code

daniel-qcode:
@PeterChaplin could you take a look at the changes below and merge if you think they are ok ?

1. added option to specify whether the payload should be downloaded or displayed in the browser.
    This allows you to either display PDF (or .csv, .txt, xlsx etc) files in a browser or have the browser give the user the option to save it to disk.
2. Changed the code to use ns_returnfile as reading binary PDFs and using ns_return didn't seem to work for me.

### DIFF
--- a/tcl/return.tcl
+++ b/tcl/return.tcl
@@ -3,54 +3,68 @@ namespace eval qc {
 }
 
 proc qc::return2client { args } {
-    #| Return data to http client
-    # Usage return2client ?code code? ?content-type mime-type? ?file file? ?filename filename? ?html html? ?text text? ?xml xml? ?json json? ?filter_cc boolean? ?header header? .. 
+    #| Return data to http client if a connection exists otherwise just output the given content.
+    # Usage return2client ?code code? ?content-type mime-type? ?html html? ?text text? ?xml xml? ?json json? ?file file? ?filename filename? ?download boolean? ?filter_cc boolean? ?header header? .. 
     set arg_names [qc::args2vars $args]
-    set headers [lexclude $arg_names file filename html xml text json code content-type]
+    set headers [lexclude $arg_names file filename html xml text json file filename download code content-type filter_cc]
     default code 200
     default filter_cc no
+    default filename [qc::url_path [ns_conn url]]
 
     if { [info exists file] } {
-        default filename [qc::url_path [ns_conn url]]
+        # File
+        default download yes
         default content-type [qc::mime_type_guess $filename]
-        default content-disposition "attachment; $filename"
-        set headers [qc::lunion $headers [list "content-disposition"]]
-        set f [open $file rb]
-        set content [read $f]
-        close $f
-        set var content
-
+        set var file
     } elseif { [info exists html] } {
+        # HTML payload
+        default download no
 	default content-type "text/html; charset=utf-8"
 	set var html
-
     } elseif { [info exists xml] } {
+        # XML payload
+        default download no
 	default content-type "text/xml; charset=utf-8"
 	set var xml
-
     } elseif { [info exists text] } {
+        # Text payload
+        default download no
 	default content-type "text/plain; charset=utf-8"
 	set var text
-
     } elseif { [info exists json] } {
+        # JSON payload
+        default download no
 	default content-type "application/json; charset=utf-8"
 	set var json
-
     } else {
-	error "No payload given in html or xml or text or json" 
+        # Unknown payload
+	error "No payload given in html or xml or text or json or file" 
     }
 
-    if { $filter_cc } {
-	# mask credit card numbers
-	set $var [qc::format_cc_masked_string [set $var]] 
+    # Update headers and return payload to client
+    if { [expr 0x1 & [ns_conn flags]] } {
+        # no open connection
+        return [lindex $args end]
+    } else {
+        if { $filter_cc } {
+            # mask credit card numbers
+            set $var [qc::format_cc_masked_string [set $var]] 
+        }
+        if { $download && "content-disposition" ni $headers } {
+            set content-disposition "attachment; filename=$filename"
+            lappend headers content-disposition       
+        }
+        # Other headers
+        foreach name $headers {
+            set conn_headers [ns_conn outputheaders]
+            ns_set update $conn_headers $name [set $name]
+        }
+        if { $var eq "file" } {
+            ns_returnfile $code ${content-type} [set $var]
+        } else {
+            ns_return $code ${content-type} [set $var]
+        }
     }
-    # Other headers
-    foreach name $headers {
-    	set conn_headers [ns_conn outputheaders]
-	ns_set update $conn_headers $name [set $name]
-    }
-
-    ns_return $code ${content-type} [set $var]
 }
 
 proc qc::return_html { html } { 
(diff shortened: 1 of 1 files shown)


PeterChaplin:
I switched away from ns_returnfile after reading the documentation about it sometimes returning the wrong file:
http://panoptic.com/wiki/aolserver/Ns_returnfile
I don't know if it's likely to be an issue.

tcl/return.tcl
```diff
@@ -3,54 +3,68 @@ namespace eval qc {
 }
 
 proc qc::return2client { args } {
-    #| Return data to http client
-    # Usage return2client ?code code? ?content-type mime-type? ?file file? ?filename filename? ?html html? ?text text? ?xml xml? ?json json? ?filter_cc boolean? ?header header? .. 
+    #| Return data to http client if a connection exists otherwise just output the given content.
```
PeterChaplin: I assume the motivation behind this is to make unit testing of procs easier?

tcl/return.tcl
```diff
@@ -3,54 +3,68 @@ namespace eval qc {
...
+    # Usage return2client ?code code? ?content-type mime-type? ?html html? ?text text? ?xml xml? ?json json? ?file file? ?filename filename? ?download boolean? ?filter_cc boolean? ?header header? .. 
     set arg_names [qc::args2vars $args]
-    set headers [lexclude $arg_names file filename html xml text json code content-type]
+    set headers [lexclude $arg_names file filename html xml text json file filename download code content-type filter_cc]
     default code 200
     default filter_cc no
+    default filename [qc::url_path [ns_conn url]]
```
PeterChaplin: Have you tested this when no connection exists? I would expect it to error if called from a cp or tclsh

PeterChaplin:
Regarding content-disposition and filenames, should we be using "inline" content-disposition to specify a filename when we aren't specifying "attachment" -
http://www.faqs.org/rfcs/rfc2183.html ?

tcl/return.tcl
```diff
@@ -3,54 +3,68 @@ namespace eval qc {
...
-	ns_set update $conn_headers $name [set $name]
+    # Update headers and return payload to client
+    if { [expr 0x1 & [ns_conn flags]] } {
+        # no open connection
+        return [lindex $args end]
+    } else {
+        if { $filter_cc } {
+            # mask credit card numbers
+            set $var [qc::format_cc_masked_string [set $var]] 
+        }
```
PeterChaplin: If `filter_cc` is specified and no connection exists, shouldn't we filter the return data?

daniel-qcode:
```
I switched away from ns_returnfile after reading the documentation about it sometimes returning the wrong file:
http://panoptic.com/wiki/aolserver/Ns_returnfile
I don't know if it's likely to be an issue.
```

This was in aolserver not naviserver so it may not be an issue anymore.

I couldn't find anything about people complaining about this issue in naviserver so I think ns_returnfile should be fine.

Also this post seems to suggest that naviserver uses a differenent caching technique that wouldn't be as  susceptible to this: http://comments.gmane.org/gmane.comp.web.naviserver.devel/2822

daniel-qcode:
```
I assume the motivation behind this is to make unit testing of procs easier?
```

No I think it was just to prevent us trying to return data to the client when the connection had already been closed

daniel-qcode:
```
Have you tested this when no connection exists? I would expect it to error if called from a cp or tclsh
```

I wouldn't expect return2client to be run in the control port.

If it was the existing ```ns_conn flags``` call would throw an error

daniel-qcode:
```
Regarding content-disposition and filenames, should we be using "inline" content-disposition to specify a filename when we aren't specifying "attachment" -
http://www.faqs.org/rfcs/rfc2183.html ?
```

Yes I think that would be fine It means that if we chose to display inline and user hit "Ctrl-s" we would be able to specify the filename using the new args.

Updated pull request

daniel-qcode:
```
If filter_cc is specified and no connection exists, shouldn't we filter the return data?
```

Don't think it matters too much, but have moved the code to filter credit card numbers.

daniel-qcode:
@PeterChaplin 

Updated pull request, could you take another look please ?

daniel-qcode:
Also I added support for returning CSVs

PeterChaplin:
Could you help me understand, under what circumstances would we be calling return2client after the connection had been closed? And if we did, why would we be capturing the return value? It seems to me that the correct behaviour in this case would be for return2client to throw an error.

tcl/return.tcl
```diff
@@ -3,54 +3,86 @@ namespace eval qc {
 }
 
 proc qc::return2client { args } {
-    #| Return data to http client
-    # Usage return2client ?code code? ?content-type mime-type? ?file file? ?filename filename? ?html html? ?text text? ?xml xml? ?json json? ?filter_cc boolean? ?header header? .. 
+    #| Return data to http client if a connection is open otherwise just output the given content.
+    # Usage return2client ?code code? ?content-type mime-type? ?html html? ?text text? ?xml xml? ?json json? ?csv csv? ?file file? ?filename filename? ?download boolean? ?filter_cc boolean? ?header header? .. 
     set arg_names [qc::args2vars $args]
-    set headers [lexclude $arg_names file filename html xml text json code content-type]
+    set headers [lexclude $arg_names file filename html xml text json csv file filename download code content-type filter_cc]
```
PeterChaplin: Duplication of "file" and "filename" in this list.

daniel-qcode:
I'm not sure, this was something introduced by Nicky for the blog project
but he can't remember why it was required.

@bvw do you know anything about why this is needed ?

On 11 February 2015 at 09:49, Peter Chaplin <notifications@github.com>
wrote:

> Could you help me understand, under what circumstances would we be calling
> return2client after the connection had been closed? And if we did, why
> would we be capturing the return value? It seems to me that the correct
> behaviour in this case would be for return2client to throw an error.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/qcode-software/qcode-tcl/pull/190#issuecomment-73855913>
> .
>



-- 
Kind Regards,
Daniel Clark
Qcode Software Limited
www.qcode.co.uk
Tel: +44(0) 1463 896 482

PeterChaplin:
I'm wondering about what defaults we should use for the `download` option - does it make sense that csv is an attachment by default, while json is inline by default?

Maybe we should *always* use "inline" as the default, and let the developer specify when they want something to be treated as an attachment.

daniel-qcode:
I not sure It is even possible to view csv and xls inline

I set the defaults to be consistant with how we currently treat csv, json,
html etc

On 11 February 2015 at 10:20, Peter Chaplin <notifications@github.com>
wrote:

> I'm wondering about what defaults we should use for the download option -
> does it make sense that csv is an attachment by default, while json is
> inline by default?
>
> Maybe we should *always* use "inline" as the default, and let the
> developer specify when they want something to be treated as an attachment.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/qcode-software/qcode-tcl/pull/190#issuecomment-73859620>
> .
>



-- 
Kind Regards,
Daniel Clark
Qcode Software Limited
www.qcode.co.uk
Tel: +44(0) 1463 896 482

bvw:
Yes a handler could take care of returning to client itself and close the
connection or just return html from the proc in which case the caller would
check if the connection was still open and return the response.

bvw:
The defaults for download are probably okay most of the time and we always have the option of making the option explicit.
@PeterChaplin , all okay with me. Please merge if nothing else outstanding.

PR #194, opened 2015-02-13
JSON Response Extend

nicky-johnstone:
Extends the JSON Response to include a `status` element that indicates `valid` or `invalid`.

Checklist
----
  - Unit testing `qc::status valid` and `qc::status invalid`.
  - Testing `qc::data2json` after setting the status and also without setting the status to test the default.


nicky-johnstone:
One thing we may want to consider is automatically setting the status to `invalid` if a record item is set as invalid. Not sure if it's best to do this or leave it to the developer to take care of.

For example:

```tcl
qc::record invalid username $username "That username is already taken."
```

would automatically call

```tcl
qc::status invalid
```

bvw:
Lets not create or rely on a functional dependency between the global status and the aggregate status of each field.
It may be conceivable for each of the 4 possible combinations to make sense at some point.

| All valid | Global Valid |
| --------- | -------------- |
| Yes | Yes |
| Yes | No |
| No | Yes |
| No | No |

bvw:
`qc::status` seems like quite a special name but just floating ideas.
Should we consider names qualified by a namespace etc.

`qc::status invalid`
`qc::response::status invalid`
`qc::response_status invalid`

`qc::record invalid username $username "That username is already taken"`
`qc::response::record invalid username $username "That username is already taken"`
`qc::response_record invalid username $username "That username is already taken"`

nicky-johnstone:
I've created the namespace `qc::response` and moved the API over to this namespace. I'd agree that it is more fitting and sensible to do so because it makes the intent of the call more obvious. Also easier to do now while the API is still young and not widely used.

The API is now called as such:

```
qc::response status x
qc::response record x
qc::response message x
qc::response action x
```

where `x` is a procedure of that particular ensemble.

daniel-qcode:
@bvw looks good to me

PR #197, opened 2015-02-19
Filter authenticate update

nicky-johnstone:
The current user ID was not being set up if the session was recognised as valid. Changed to set the current user ID.

`qc::auth_session` changed to check validity of session rather than existence.

#### Checklist
----
  - Tested change on blog project. All working as expected.

### DIFF
--- a/tcl/auth.tcl
+++ b/tcl/auth.tcl
@@ -15,8 +15,9 @@ proc qc::auth {} {
     if { [qc::cookie_exists session_id] } {
         set session_id [cookie_get session_id]
         if { [qc::session_exists $session_id] } {
+            set current_user_id [qc::auth_session $session_id]
             qc::session_update $session_id
-            return [set current_user_id [qc::auth_session $session_id]]
+            return $current_user_id
         }
     }
     
@@ -72,7 +73,7 @@ proc qc::auth_hba_check {} {
     } 
 }
 
-proc qc::auth_session { session_id } {
+proc qc::auth_session { session_id} {
     #| Try to authenticate user based on the session_id given
     #| Return the user_id if successful
     #| On failure throw AUTH error
--- a/tcl/auth.tcl
+++ b/tcl/auth.tcl
@@ -14,7 +14,7 @@ proc qc::auth {} {
     # Try session based auth
     if { [qc::cookie_exists session_id] } {
         set session_id [cookie_get session_id]
-        if { [qc::session_valid $session_id] } {
+        if { [qc::session_exists $session_id] } {
             qc::session_update $session_id
             return [set current_user_id [qc::auth_session $session_id]]
         }
@@ -34,7 +34,7 @@ proc qc::auth_check {} {
     # session based auth
     if { [qc::cookie_exists session_id]} {
 	set session_id [cookie_get session_id]
-	if { [qc::session_valid $session_id] } {
+	if { [qc::session_exists $session_id] } {
 	    return true
 	}
     }
@@ -76,7 +76,7 @@ proc qc::auth_session { session_id } {
     #| Try to authenticate user based on the session_id given
     #| Return the user_id if successful
     #| On failure throw AUTH error
-    if { [session_exists $session_id] } {
+    if { [session_valid $session_id] } {
         return [session_user_id $session_id]
     } else {
 	error "Session authentication failed to identify you." {} AUTH
--- a/tcl/auth.tcl
+++ b/tcl/auth.tcl
@@ -14,7 +14,7 @@ proc qc::auth {} {
     # Try session based auth
     if { [qc::cookie_exists session_id] } {
         set session_id [cookie_get session_id]
-        if { [qc::session_exists $session_id] } {
+        if { [qc::session_valid $session_id] } {
             qc::session_update $session_id
             return [set current_user_id [qc::auth_session $session_id]]
         }
@@ -34,7 +34,7 @@ proc qc::auth_check {} {
     # session based auth
     if { [qc::cookie_exists session_id]} {
 	set session_id [cookie_get session_id]
-	if { [qc::session_exists $session_id] } {
+	if { [qc::session_valid $session_id] } {
 	    return true
 	}
     }
--- a/tcl/filter.tcl
+++ b/tcl/filter.tcl
@@ -50,13 +50,10 @@ proc qc::filter_authenticate {event {error_handler qc::error_handler}} {
         # Check if this request is registered.
         if { [qc::registered $method $url_path] } {
             if {[qc::cookie_exists session_id] && [qc::session_valid [qc::session_id]]} {
-                global session_id current_user_id
-                qc::session_update $session_id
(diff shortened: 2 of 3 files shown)


tcl/auth.tcl
```diff
@@ -76,7 +76,7 @@ proc qc::auth_session { session_id } {
     #| Try to authenticate user based on the session_id given
     #| Return the user_id if successful
     #| On failure throw AUTH error
-    if { [session_exists $session_id] } {
+    if { [session_valid $session_id] } {
         return [session_user_id $session_id]
```
daniel-qcode: @bvw Is this what we want this will enforce a default age limit of 12 hours and idle_timeout of 1 hour for all projects, even TLC ?

nicky-johnstone: Likely not what is wanted for everything. However simply checking existence won't work because the session could be expired.

Not sure how we can get around this without having to change all calls to `auth_session`.

tcl/filter.tcl
```diff
@@ -50,10 +50,13 @@ proc qc::filter_authenticate {event {error_handler qc::error_handler}} {
         # Check if this request is registered.
         if { [qc::registered $method $url_path] } {
             if {[qc::cookie_exists session_id] && [qc::session_valid [qc::session_id]]} {
-                qc::session_update [qc::session_id]
+                global session_id current_user_id
+                qc::session_update $session_id
+                set current_user_id [qc::session_user_id $session_id]
             } elseif {$method ni [list "GET" "HEAD"] && [qc::cookie_exists session_id] && ! [qc::session_valid [qc::session_id]]} {
```
daniel-qcode: I think I'd prefer not to expose the global session_id here.

qc::session_id should be optomised to use the global session_id variable itself so I'd prefer to call ```qc::session_id``` multiple times or use a local variable to store it temporarily ```set session_id [qc::session_id]```

daniel-qcode: Is setting the current_user_id necessary here, would have thought the first time qc::auth is called this would be handled for us. Or am I missing something ?

tcl/filter.tcl
```diff
@@ -84,7 +87,9 @@ proc qc::filter_authenticate {event {error_handler qc::error_handler}} {
             
             # Roll the anonymous session after 1 hour.
             if {[qc::auth] == [qc::anonymous_user_id]} {
-                qc::cookie_set session_id [qc::anonymous_session_id]
+                global session_id
+                set session_id [qc::anonymous_session_id]
+                qc::cookie_set session_id $session_id
             }
```
daniel-qcode: This is fine just mentioning that another alternative would be to unset the global session_id variable and let subsequent calles to ```qc::session_id``` determine the session id directly from the recently set cookie.
```
qc::cookie_set session_id [qc::anonymous_session_id]
unset ::session_id
```

tcl/auth.tcl
bvw: Can the definition of age and idle_timeout be config parameters with defaults if not defined.

tcl/filter.tcl
bvw: Yes, manipulation of the global `current_user_id` should only be done via an API where possible.

nicky-johnstone: Calling `qc::auth` without having `current_user_id` set would run through authentication again. That's probably fine though as it likely wouldn't impact performance to any significant degree. `qc::auth` wasn't used here because we didn't want HBA for this filter although I don't recall why.

I changed it for consistency between this section and implicitly logging in the anonymous user which does set the current user ID. Now that I've looked through it setting `current_user_id` anywhere outside of `qc::auth` probably isn't necessary but the use for logging in the anonymous session had me believing that it was.

tcl/filter.tcl
bvw: Not sure this will work because cookie_set only sets outbound reply http headers while cookie_get reads received headers.

tcl/auth.tcl
nicky-johnstone: I've also just realised something odd with `qc::auth` unless I am misunderstanding the intentions:

```tcl
# Try session based auth
    if { [qc::cookie_exists session_id] } {
        set session_id [cookie_get session_id]
        if { [qc::session_exists $session_id] } {
            qc::session_update $session_id
            return [set current_user_id [qc::auth_session $session_id]]
        }
    }
```

It looks to me like it would update the session and then check it which would mean that the idle timeout for a session would never fail because the time-modified was just updated and won't reflect the actual idle time.

I think it should be:

```tcl
# Try session based auth
    if { [qc::cookie_exists session_id] } {
        set session_id [cookie_get session_id]
        if { [qc::session_exists $session_id] } {
            set current_user_id [qc::auth_session $session_id]
            qc::session_update $session_id
            return $current_user_id
        }
    }
```

bvw: `session_exists`
`session_valid`

Amazon have this model on their website.
A returning customer whom we recognise but we need to check if his session is valid before allowing more sensitive operations.

However the purpose of authentication is normally to check for permission to authorise the user to perform an action.
We could could call these "recognising" a user and "authenticating" as user.

The code as it stands in the lib doesn't make use of session_valid so doesn't go there with regards to validity.

May require `session_valid` to be used in `auth_session` rather than session_exists ?

Any ramifications for TLC checkout ?

daniel-qcode: I think currently sessions for tlc-erp and mla-erp are non-expiring.

So the current behavior in qcode-tcl is that we only check that a session exists not whether it is valid according to timeouts and age limits.

For mla-erp sessions are expired every night by deleting all sessions from the database.

I think tlc-erp sessions never expire.

The behavior for expiring sessions was introduced of tlc-www and mla-www as custom authentication procs which is why this behavior has not made it into qcode-tcl yet.

So in short I think what you suggested is probably the desired behavior going forward.

@ bvw anything to add

daniel-qcode: If we make the change to ```qc::auth_session``` there will be ramifications for projects like tlc-www & tlc-erp but if we allow the developer to override ```qc::auth_session```'s default ```age_limit``` and ```idle_timeout``` I think that would be ok.

something like:

```
proc qc::session_valid {args} {
    #| Check session exists and has not expired.
    qc::args $args -age_limit ? -idle_timeout "1 hour" -- session_id

    if { [qc::param_exists session_age_limit] } {
        default age_limit [qc::param_get session_age_limit]
    } else {
        deafult age_limit "12 hours"
    }
    if { [qc::param_exists session_idle_timeout] } {
        default idle_timeout [qc::param_get session_idle_timeout]
    } else {
        deafult idle_timeout "1 hour"
    }

    set qry {
        select
        user_id
        from session
        where
        session_id=:session_id
        and (current_timestamp-time_modified) <=:idle_timeout::interval
        and (current_timestamp-time_created) <=:age_limit::interval
    }

    db_0or1row $qry {
        return false
    } {
        return true
    }
}
```

bvw: ok lets try that.

PR #199, opened 2015-02-23
bug fix: parameter attribute names are case insensitive

daniel-qcode:


daniel-qcode:
This is to resolve the bugs in support@qcode.co.uk this morning.

There was a email with parameter attribute "Boundary" and we were only matching "boundary".

According to RFC 2045 parameter attribute names are case insenstive:

```
parameter := attribute "=" value

attribute := token
    ; Matching of attributes
    ; is ALWAYS case-insensitive.
```

PR #207, opened 2015-02-27
EOD commit

daniel-qcode:
NOT ready for merge

### DIFF
--- a/tcl/filter.tcl
+++ b/tcl/filter.tcl
@@ -58,7 +58,7 @@ proc qc::filter_authenticate {event args} {
             }
 
             if { $method in [list "GET" "HEAD"] && [qc::session_user_id [qc::session_id]] == [qc::anonymous_user_id] && [qc::session_id] != [qc::anonymous_session_id] } {
-                # GET/HEAD request for anonomous session - roll session after 1 hour
+                # GET/HEAD request for anonymous session - roll session after 1 hour
                 global session_id current_user_id
                 set current_user_id [qc::anonymous_user_id]
                 set session_id [qc::anonymous_session_id]
--- a/tcl/filter.tcl
+++ b/tcl/filter.tcl
@@ -38,38 +38,71 @@ proc qc::filter_validate {event {error_handler qc::error_handler}} {
     }    
 }
 
-proc qc::filter_authenticate {event {error_handler qc::error_handler}} {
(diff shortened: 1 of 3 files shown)


daniel-qcode:
Can we get Nicky to review and test these changes for mla-www ?

I tried checking out mla-www's new-transaction-implementation branch but schema_update won't apply cleanly

tcl/filter.tcl
```diff
@@ -38,38 +38,71 @@ proc qc::filter_validate {event {error_handler qc::error_handler}} {
...
+            if { ![qc::cookie_exists session_id] || ![qc::session_exists [qc::session_id]] } {
+                # No session cookie or session doesn't exist - implicitly log in as anonymous user.
+                global session_id current_user_id
+                set current_user_id [qc::anonymous_user_id]          
+                set session_id [qc::anonymous_session_id]
+                qc::cookie_set session_id $session_id
+            }
+
+            if { $method in [list "GET" "HEAD"] && [qc::session_user_id [qc::session_id]] == [qc::anonymous_user_id] && [qc::session_id] != [qc::anonymous_session_id] } {
+                # GET/HEAD request for anonomous session - roll session after 1 hour
```
nicky-johnstone: Typo - anonomous

nicky-johnstone:
I seem to be getting redirected after being redirected to `/session-expired`.

image

Here's what I've done:

1) Log in.
2) In the DB make the session invalid (I just set date created to be 3 days in the past).
3) Attempt to navigate to another page.

So after this I get directed to `/session-expired` as expected however it would seem that the redirect to session expired pages gets redirected once more. This completely throws off the `next_url` form variable so when you do sign back in the `next_url` is the session expired page.

It does not appear to happen when navigating to the home page in step 3.

I've not looked too much into it and it could be an obvious problem that's on my end rather than the filter but my head can't navigate the code properly at the moment due to trying to understand shenanigans with some other bugs and the authentication query problem. :sleeping:

nicky-johnstone:
Working as far as I've been able to test with the changes to mla-www.

The bug mentioned in my comment above was due to the fact that not all URLs have been shifted over to the new transaction process and therefore did not pass through `filter_authenticate` initially. It should redirect the redirect once the other handlers have been shifted over to the new transaction process.

PR #212, opened 2015-03-17
Use column instead of name for returned record response.

nicky-johnstone:
`$name` is the name of the argument for the handler that may or may not be fully qualified i.e. `table.column` or `column`. The record was returned with the potentially fully qualified name of the argument which does not match up with the form elements and therefore was causing troubles with the validation plugin because it cannot select the input elements with a fully qualified name. The response should just return the column portion of a fully qualified name and this change fixes that.

The alternative is to have the validation plugin check if the name is fully qualified and split name by the separator `.`.

### DIFF
--- a/tcl/validate.tcl
+++ b/tcl/validate.tcl
@@ -16,11 +16,11 @@ proc qc::validate2model {dict} {
         
         # Check if nullable
         if {! $nullable && $value eq ""} {
-            qc::response record invalid $name $value $message
+            qc::response record invalid $column $value $message
             set all_valid false
             continue
         } elseif {$nullable && $value eq ""} {
-            qc::response record valid $name $value ""
+            qc::response record valid $column $value ""
             continue
         }
         # Check value against data type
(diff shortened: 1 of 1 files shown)


bvw:
Can you please give example form element names and example table column names to help me understand.

nicky-johnstone:
A good example is the one I was working on to encounter the error:

```
register POST /catalogue-request {catalogue_request.firstname ...} {
     #| Handler for a catalogue request`
     ...
}
```

This handler has lots of args but they are common in the database. Therefore to handle them appropriately I had to qualify the args with the table name.

`validate2model` would use the fully qualified name to set up the record:

```json
"record": {
    "catalogue_request.firstname": {
        
    }
}
```

The validation plugin would try to select the form element named `catalogue_request.firstname` but in CSS selectors this is an element `catalogue_request` with class `firstname`.

Form elements are usually named for the column in the database that the data represents. Using just the column portion as the name for the record fixes the above error so that the form element can be selected.

```json
"record": {
    "firstname": {
        
    }
}
```

PR #219, opened 2015-04-02
Bug fix cast decimal

daniel-qcode:


daniel-qcode:
This is to make the following calls have the same result:

```
$ cast decimal 001
001.123
$ cast decimal -precision 10 001
1
```

After this change leading zeros will be stripped out and both will return ```1```

PR #226, opened 2015-04-09
Db cache size param2

daniel-qcode:


tcl/db_cache.tcl
```diff
@@ -141,7 +141,24 @@ proc qc::db_cache_select_table { args } {
         # Use ns_cache
         # Create the cache if it doesn't exist yet
         if { ! [in [ns_cache_names] db] } {
-	    ns_cache create db -size [expr 1024*1024] 
+            # DB cache size
+            set param_name db_cache_size
+            db_0or1row {select param_value from params where param_name=:param_name} {
```
bvw: My concern is that this qry is run for every request which has a performance overhead.
Also we need to be able to run qcode outside Naviserver for muppet where we have no db access.

daniel-qcode: It is only run once when the db ns_cache hasn't been initialised

daniel-qcode: ns_cache is specific to naviserver anyway

bvw: ok fair enough

PR #227, opened 2015-04-09
alternative just using Naviserver param

bvw:


tcl/db_cache.tcl
```diff
@@ -141,20 +141,15 @@ proc qc::db_cache_select_table { args } {
...
-                    set db_cache_size $param_value
-                } else {
-                    # use default of 10 MB
-                    set db_cache_size [expr 1024*1024]
-                }
-            } {
-                # DB param exists 
-                set db_cache_size as param_value
+            if { [info commands ns_config] eq "ns_config" && \
+                     [set param_value [ns_config ns/server/[ns_info server] $param_name]] ne "" } {
```
daniel-qcode: I didn't bother checking this condition because ns_cache is specific to the code running in timeserver anyway so ns_config command should always exist

PR #236, opened 2015-06-01
Db catch error globals update

nicky-johnstone:


tcl/db.tcl
```diff
@@ -387,7 +387,11 @@ proc qc::db_trans {args} {
 	    } else {
 		incr db_trans_level($db) -1
 	    }
-	    return -code $code $result
+            
+            if { $return_code == 2 } {
```
nicky-johnstone: If we wanted to support `return -code x` statements I think that updating this code to:

```tcl
if { $return_code == 2 && [dict get $options -code] == 0 } {
    ...
}
```

would allow us to do so. I'd need to test more cases though to make sure they'd all be covered. Not entirely sure if that support is necessary at the moment.

bvw:
Could we list what testing has been done ie test code.

nicky-johnstone:
I've tried variations on the following code:

```tcl
proc test_try {} {
    try {
        error "FOO"
    } {
        log "BAR"
    }
    log "myFoo = $myFoo"
    log "FINISH test_try2"
}

proc wrapper {} {
    ::try {
        test_try
    } on error [list error_message options] {
        log "ERROR: $error_message"
    }
}

proc test_db_foreach {{limit 10}} {
    db_foreach {
        select
        user_code
        from users
        order by user_id
        limit :limit
    } {
        if { $user_code == "ROOT" } {
            break
        }
        log $user_code   
    } {
        error "No results"
    }
}

proc test_db_0or1row {{user_id 0}} {
    db_0or1row {
        select
        user_code
        from
        users
        where user_id = :user_id
    } {
        error "User not found"
    } {
        return "FOUND: $user_code"
    }
    log "FINISH test_db_0or1row"
}


proc test_db_cache_foreach {{limit 10}} {
    db_cache_foreach {
        select
        user_code
        from users
        order by user_id
        limit :limit
    } {
        if { $user_code == "ROOT" } {
            break
        }
        log $user_code
        
    } {
        error "No results"
    }
}

proc test_db_cache_0or1row {{user_id 0}} {
    db_cache_0or1row {
        select
        user_code
        from
        users
        where user_id = :user_id
    } {
        error "User not found"
    } {
        return "FOUND: $user_code"
    }
    log "FINISH test_db_cache_0or1row"
}
```

Variations:
* Changing `break` for `continue` on the foreach tests.
* Changing errors for returns and vice versa.

tcl/db.tcl
```diff
@@ -387,7 +387,13 @@ proc qc::db_trans {args} {
 	    } else {
 		incr db_trans_level($db) -1
 	    }
-	    return -code $code $result
+            
+            if { $return_code == 2 && [dict get $options -code] == 0 } {
+                dict set options -code return
+            } else {
+                dict incr options -level
```
nicky-johnstone: I've added this increment to level because I realised that under most circumstances, since we're executing code in the caller's stack frame, then we want to return values while taking into account the caller's stack frame.

I'll add another comment shortly with the testing performed and attempt to explain why the level should be incremented here.

nicky-johnstone:
## Return codes and levels
Looking at examples of return codes and what they mean is hopefully helpful here.

Bear in mind that the default `-level` for `return -code x` is `1`. I.e. `return -code error` is the same as `return -code error -level 1`.

```
error "FOO"                          ->    Throw error now
return -code error "FOO"             ->    Throw error in caller
return -code error -level 2 "FOO"    ->    Throw error in caller of caller


return "FOO"                         ->    Return now
return -code return "FOO"            ->    Return up to, and including, caller
return -code return -level 2 "FOO"   ->    Return up to, and including, caller of caller
```

Some test cases to confirm:

```tcl
proc test {} {
    log "START: test"
    return "FOO"
    log "FINISH: test"
}

proc caller_of_test {} {
    log "START: caller_of_test"
    test
    log "FINISH: caller_of_test"
}

proc caller_of_caller_of_test {} {
    log "START: caller_of_caller_of_test"
    caller_of_test
    log "FINISH: caller_of_caller_of_test"
}
```

If we call `caller_of_caller_of_test` then we'd expect to see all log statements except `FINISH: test` because it's after the return statement and therefore should never be executed.
```
> caller_of_caller_of_test

START: caller_of_caller_of_test
START: caller_of_test
START: test
FINISH: caller_of_test
FINISH: caller_of_caller_of_test
```

If we change the return statement in `foo` to use level 1...

```tcl
proc test {} {
    log "START: test"
    return -code 2 "FOO"
    log "FINISH: test"
}
```

```
> caller_of_caller_of_test

START: caller_of_caller_of_test
START: caller_of_test
START: test
FINISH: caller_of_caller_of_test
```

We don't get the `FINISH: caller_of_test` log output because `test` has told the caller to return. Therefore any statements after the call to `test` in the caller will not be executed.

And if we modify `test` to level 2...

```tcl
proc test {} {
    log "START: test"
    return -code 2 -level 2 "FOO"
    log "FINISH: test"
}
```

```
> caller_of_caller_of_test

START: caller_of_caller_of_test
START: caller_of_test
START: test
```

We no longer have the log output `FINISH: caller_of_caller_of_test` because `test` has told it to return by specifying `-level 2`.

## Test cases for code changes

Changes to the 0 rows code in `foo` is what I had been testing with various return codes and levels.

The test is initiated by calling `baz`.

```tcl
proc foo {} {
    ::try {
        db_0or1row {
            select user_code from users where user_id = -2
        } {
            # 0 rows
            <test case code>
        } {
            return $user_code
        }
    } on error [list error_message options] {
        log "TRY 1"
    }
}


proc bar {} {
    ::try {
        foo
        log "BAR: AFTER FOO"
    } on error [list error_message options] {
        log "TRY 2"
    }
}

proc baz {} {
    ::try {
        bar
        log "BAZ: AFTER BAR"
    } on error [list error_message options] {
        log "TRY 3"
    }
}
```

### Test Case: `error "FOO"`

**Expected output:**

Throw the error in `foo`
```
TRY 1
BAR: AFTER FOO
BAZ: AFTER BAR
```

**Actual output before level increment change:**

```
TRY 1
BAR: AFTER FOO
BAZ: AFTER BAR
```

**Actual output after level increment change:**

```
TRY 1
BAR: AFTER FOO
BAZ: AFTER BAR
```


### Test Case: `return -code error "FOO"`

**Expected output:**

Throw the error in `bar`
```
TRY 2
BAZ: AFTER BAR
```

**Actual output before level increment change:**

```
TRY 1
BAR: AFTER FOO
BAZ: AFTER BAR
```
**Note**: The error was thrown inside `foo` instead of being thrown in the caller of `foo`.

**Actual output after level increment change:**

```
TRY 2
BAZ: AFTER BAR
```

### Test Case: `return -code error -level 2 "FOO"`

**Expected output:**

Throw the error in `baz`
```
TRY 3
```

**Actual output before level increment change:**

```
TRY 2
BAZ: AFTER BAR
```
**Note**: The error was thrown inside `bar` instead of being thrown in `baz`.

**Actual output after level increment change:**

```
TRY 3
```

nicky-johnstone:
Hopefully the examples and test cases above are clear but if not then let me know.

I've still to run test cases for the other changed procedures as well as other return codes just to ensure that it's all going to behave as expected.

nicky-johnstone:
I believe there are further special cases for `break` and `continue` regarding level increment because they have to be used within a loop construct which in the case of `db_foreach` and `db_cache_foreach` you would want to break or continue at level 0 (i.e. within `db_foreach` or `db_cache_foreach` and not in the caller`).

Looking into this and testing.

nicky-johnstone:
I've just realised that my view on returns within DB scripts might be wrong. I've been expecting DB scripts to act in much the same way as `::try` although this could be wrong.

I'm wondering what would be the expectations and what should we do for things like;

```tcl
proc test {} {
    db_foreach {
        # SQL
    } {
        return -code error -level 1 "Message"
    } 
    
    # other code
}
```

Would you expect `test` to throw the error? Or would you expect the caller of test to throw the error?

nicky-johnstone:
An example using return. Perhaps this one is more clear than the `error` example above.

```tcl
proc test {} {
    db_foreach {
        # SQL
    } {
        return "FOO"
    }
    
    # other code
}
```

Would the above code be expected to return out of the procedure `test` with value `FOO`?

Or would it be expected that `# other code` is executed?

nicky-johnstone:
Updated and ready for merge.

I've tested the updated procs individually (except for `qc::lock`) with varying kinds of returns and levels. I believe they all now function as expected as per the discussion we had yesterday.

bvw:
I've also done some testing on this and all works as expected.

Do you have pgTcl to run the test suite ?

nicky-johnstone:
I don't. Is `libpgtcl` in the repository the one I should install or are we after a different version?

bvw:
yes I had to search here:-https://packages.debian.org/source/squeeze/database/pgtcl
we need
`sudo apt-get install libpgtcl1.5`
perhaps also the dev package
`sudo apt-get install libpgtcl-dev`

nicky-johnstone:
All current tests pass with three exceptions that aren't related to these changes.

At some point we'll also need to add tests for the DB introspection stuff that was added as part of the new transaction process. And perhaps add some more test cases for different return types used within the recently changed DB procs.

nicky-johnstone:
The tests that fail:

* `db_connect` seems to return a string.
  * Can be fixed by setting up the result for this test case.

```
==== db_connect-1.0 db_connect FAILED
==== Contents of test case:

            db_connect {*}$test_conf
        
---- Result was:
pgsql7
---- Result should have been (exact matching):

==== db_connect-1.0 FAILED

```


* `db_col_varchar_length` fails due to `ns_cache_names` being an invalid command.
  * About a month ago we changed the DB introspection stuff to make use of caching so this is why it's now failing.

* `form_db` fails because it cannot get an authenticity token from the current session.

Do we simply remove the latter two test cases or is there a way of feasibly continuing to test them with some setup?

nicky-johnstone:
## Tests
- [x] `db_connect`
  - Removed
- [x] `db_col_varchar_length`
  - `db_cache_select_table` updated to check for existence of `ns_cache` proc to allow for caching when Naviserver isn't present.
- [ ] `fom_db`
  - Requires session to be present so that authenticity token relating to current session can be used in the form to aid against CSRF particularly for forms of method `POST`.
  - Options:
      - Exclude this test.
      - Update `form_db` to check `qc::conn_open` for a connection and add authenticity token if connection open.
      - Update `form_db` to check `method` of form. If `POST` then include authenticity token otherwise don't.
          - This would mean that we cannot test `form_db` with method `POST` but can for any other method.

PR #247, opened 2015-07-15
Always log in the anonymous user in no session

nicky-johnstone:
[MLA issue #832](https://github.com/qcode-software/mla-www/issues/832#issuecomment-121573933)


nicky-johnstone:
Appears to be working from my testing on MLA WWW and ERP.

PR #253, opened 2015-07-28
Single colon for port

nicky-johnstone:
@PeterChaplin Looks like you were using double colon for the port delimiter when it should be just a single colon. Think I've caught them all but could you have a look over to make sure?


PeterChaplin:
Looks good to me. Do you want me to merge?

bvw:
yes please if you are happy with it.

PR #254, opened 2015-07-28
Correction to response extend usage error

PeterChaplin:


nicky-johnstone:
Usage can be either:

1. `Usage: qc::response extend name key value ?key value ...?`
2. `Usage: qc::response $name key value ?key value ...?`

e.g.

1. `qc::response extend foo a 1 b 2 c 3`
2. `qc::response foo a 1 b 2 c 3`

Given that we can't determine at that stage how the user called the proc then being more explicit with the usage error is probably best.

PR #258, opened 2015-08-03
Error on key missing or colon var value begins with colon

nicky-johnstone:
Tcl
----
  - URL test suite with new tests written for colon variable errors.
  - Browsing some elements of mla-erp known to use URL e.g. product select table, report drill downs...


tcl/url.tcl
```diff
@@ -24,11 +25,15 @@ proc qc::url { url args } {
             set segment [string range $segment 1 end]
             # check if caller has provided a substitution for the segment
             if { [dict exists $dict $segment] } {
-                lappend substituted_segments [dict get $dict $segment]
+                set sub [dict get $dict $segment]
```
bvw: Would prefer the variable name `$value` if it doesn't conflict.

nicky-johnstone:
Updated.

PR #264, opened 2015-08-06
Bug fixes

PeterChaplin:


daniel-qcode:
@bvw This looks reasonable to me ... can you take a look over this too before I merge ?

bvw:
This interpretation of the query string is reasonable but at some point I'd like to consider the concept of a well-formed url or well-formed query string and validate against that.
Could also cast an url to a well-formed url with some information loss.
I can't think of a situation where we would want to cooperate with urls that are not well formed.
Any examples ?

PR #266, opened 2015-08-20
cast_url procs

PeterChaplin:


tcl/cast.tcl
```diff
@@ -79,6 +79,14 @@ proc qc::cast_period {string} {
...
 }
 
+proc qc::cast_url {string} {
+    return [qc::cast url $string]
+}
+
+proc qc::cast_relative_url {string} {
+    return [qc::cast relative_url $string]
+}
+
```
nicky-johnstone: `#|` comments.

If I understand correctly then these `qc::cast_` procs are required for `qc::checks` to function. If so then the `#|` should also denote these procs as deprecated since they will be superceded by the ensemble procs once current projects are fully migrated.

PeterChaplin:
Updated. Yes, the `qc::cast_*` procs were to get `qc::check` and `qc::checks` working.

tcl/cast.tcl
```diff
@@ -655,4 +663,24 @@ namespace eval qc::cast {
...
+        return -code error -errorcode CAST "Could not cast $string to an url."
+    }
+
+    proc relative_url {string} {
+        #| Cast the given string to a relative url
+        set lower [string tolower $string]
+        if { [qc::is url -relative $lower] } {
+            return $lower
+        }        
+    }
```
nicky-johnstone: Should this proc return a `CAST` error if the string isn't a relative url? Otherwise all strings will be appear to be castable to relative urls with `qc::castable`.

PeterChaplin:
Updated

PR #269, opened 2015-08-25
Trim form vars automatically

PeterChaplin:


bvw:
Could you paste in details of testing please.

bvw:
You could make the package locally and then check form submission for some test cases.
1. Simple case 
2. With whitespace
3. Repeated form variable `foo=1&foo=2`
4. Repeated form variable with whitespace.

PeterChaplin:
I've tried a make-install locally, followed by a product search for " TEST ", which matched the product codes "TEST" and "TESTAAA" on my dev machine.

I'll do some more testing after lunch.

PeterChaplin:
Created proc:
```
proc /test.html {foo} {
    return "A${foo}A"
}
```
Tested with

- `/test.html?foo=bar`
- `/test.html?foo=++bar++`
- `/test.html?foo=bar&foo=baz`
- `/test.html?foo=++bar++&foo=++baz++`

All returned as expected.

PR #270, opened 2015-08-25
Centralise authenticity token

nicky-johnstone:
Tcl
----

  - Tested with MLA ERP with header, without header, and with both header and form variable.

### DIFF
--- a/tcl/session.tcl
+++ b/tcl/session.tcl
@@ -169,3 +169,26 @@ proc qc::anonymous_user_id {} {
     return -1
 }
 
+proc qc::login {user_id} {
+    #| Logs the given user into the system.
+    global session_id current_user_id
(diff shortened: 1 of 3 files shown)


tcl/session.tcl
```diff
@@ -169,3 +169,26 @@ proc qc::anonymous_user_id {} {
     return -1
 }
 
+proc qc::login {user_id} {
+    #| Logs the given user into the system.
```
nicky-johnstone: Added this proc as I think it would prove helpful and save developers from having to know to set up the session and authenticity token cookies as well as the global session users IDs.

Having said that I'm not sure logging in a user happens in many places within a project so this abstraction may prove unnecessary due to low usage.

nicky-johnstone:
Updated.

PR #274, opened 2015-08-28
Filter validate get update

nicky-johnstone:
Tcl
----
  - Tested on MLA ERP DEV:
  - `/sales_order_select.html` form - entering invalid and valid data.
  - Entering invalid values for query string params in URLs i.e. `/sales_order.html?sales_order_id=foo`

### DIFF
--- a/tcl/html.tcl
+++ b/tcl/html.tcl
@@ -870,8 +870,7 @@ proc qc::safe_attributes {} {
     return $safe_attributes
 }
 
-
-proc qc::data2html {} {
+proc qc::global_response2html {} {
     #| Converts the global data structure to HTML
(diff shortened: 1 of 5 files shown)


tcl/html.tcl
```diff
@@ -869,3 +869,8 @@ proc qc::safe_attributes {} {
 
     return $safe_attributes
 }
+
+proc qc::global_response2html {} {
```
bvw: what about `qc::response2html`

nicky-johnstone: Considered that but didn't know if it was a little bit too generic like `data2html`.

nicky-johnstone:
Updated.

nicky-johnstone:
Updated.

PR #282, opened 2015-10-13
Expose ttl option in param_get

davidqc:


bvw:
Should we have this feature in qcode-6 ?

davidqc:
qcode-tcl-5 to 6 is a breaking upgrade I believe... but is there is an argument that we should keep the interface as consistent as we can?

In that vein I should update html2pdf in qcode-tcl-6 to match the changes in 5 also.

PR #290, opened 2015-11-18
Args length check and duplicate keys check

PeterChaplin:


bvw:
Looks good. Have you tested an even and odd number of arguments ?

PeterChaplin:
Yes, tested with:
```
% qc::args2dict [list foo bar]
foo bar
% qc::args2dict [list foo bar baz]
Missing name or value in args for qc::args2dict
% qc::args2dict [list foo bar foo baz]
Duplicate key for qc::args2dict
```

PR #298, opened 2015-12-01
Revert "Allow qc::url to pick up local vars"

bvw:
Reverts qcode-software/qcode-tcl#297

Introduces an url injection attack.


bvw:
proc /url_test.html {next_url} {
    set password secret
    return [url $next_url foo bar]
}

image

PeterChaplin:
Even with this reverted, it would still be possible for someone to call
`/url_test.html>next_url=test/:foo`
to get
`/test/bar`
which isn't a major vulnerability, since "bar" would have been exposed via the query string anyway, but we should probably look at moving to something like
`return [url :next_url foo bar]`
so as not to mix instructions with data.

PR #309, opened 2016-01-11
Update docs

daniel-qcode:


doc/filters.md
```diff
@@ -40,7 +40,7 @@ The task of this filter is to check if the session and authentication tokens are
 
 When this filter is called it will check if the request has been [registered]. If so then a check is done to determine if the session is valid.
 
-If the session is valid then the request method is checked. If it's not a GET or HEAD request the authenticity token must be checked for validity. If the token is invalid then an error is returned to the client.
+If the session is valid then the request method is checked. If it's not a GET or HEAD request the authenticity token must be checked for validity. If the token is invalid then an error is returned to the client, they are logged in as the anonymous user and they are asked to resubmit the form.
```
nicky-johnstone: This isn't currently the case as the filter has remained unchanged in the [response to an invalid authenticity token](https://github.com/qcode-software/qcode-tcl/blob/master/tcl/filter.tcl#L149).

I believe the code should remain as is but the existing documentation is misleading because the response to the client is entirely determined by the error handler chosen for the filter (which can be user-defined). The default error handler behaviour is to [return an authentication error code and message](https://github.com/qcode-software/qcode-tcl/blob/master/tcl/error.tcl#L79). If the default behaviour should be to ask the user to resubmit then we'll need to update it.

daniel-qcode: Sorry I have a custom filter_authenticate for tlc-www ... I will revert this change to docs .

Probably will be a good idea to do more testing of tlc-www filter_authenticate before back porting the change to qc::filter_authenticate

daniel-qcode:
Updated

PR #314, opened 2016-01-15
don't echo back values of sensitive fields to clients

daniel-qcode:
for discussion

### DIFF
--- a/tcl/response.tcl
+++ b/tcl/response.tcl
@@ -171,6 +171,7 @@ proc qc::response2tson {} {
     set action_objects [list object]
     set message_objects [list object]
     set extensions [list]
+
     # Default status is "valid"
     set status "valid"
     ::try {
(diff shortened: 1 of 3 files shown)


tcl/response.tcl
```diff
@@ -178,7 +179,26 @@ proc qc::response2tson {} {
...
                 record {
                     foreach {name values} $value {
-                        lappend record_objects $name [list object valid [dict get $values valid] value [list string [dict get $values value]] message [list string [dict get $values message]]]
+                        # Check if this is a sensitive column (should not be echo'ing back the value to the client)
+                        set sensitive_column false
+                        if { [qc::db_column_exists $name] } {
+                            set column $name
+                            set tables [qc::db_column_table $name]
+                            foreach table $tables {
+                                if { [qc::db_column_type $table $column] in [list password card_number] } {
```
bvw: Might be worth looking at a param for this as I can think of a few others:-
`password repeat_password password2 cc_no card_number`

bvw: perhaps a helper returning a boolean ? `qc::db_column_sensitive $table $column`

daniel-qcode: I think the same ````password``` and ```card_number``` (domain) can still be applied to those too.

Might need to do this in the future though if we come across any more ... or I can make that provision now if you like

bvw: Ah yes sorry I've confused name and type

PR #315, opened 2016-01-15
alternative approach

daniel-qcode:


tcl/response.tcl
```diff
@@ -55,21 +55,23 @@ namespace eval qc::response {
...
             #| Adds the given field to the record as invalid. If the field already exists then updates it.
             #| Also sets the status of the response to invalid.
             global data
             dict set data record $name valid false
             dict set data record $name value $value
-            dict set data record $name message $message
+            dict set data record $name message $message            
+            dict set data record $name sensitive $sensitive
             qc::response status invalid
         }
```
daniel-qcode: if validate2model flags a field as sensitive by doing:
```
qc::response record valid password secret "Password Valid" true
```

what if there were custom validation handlers like something to check password complexity and they did:
```
qc::response record invalid password secret "Password not complex enough"
```

The field would no longer be flagged as sensitive.
======================================

Of course the developer could have done the following (but he might not remember):
```
qc::response record invalid password secret "Password not complex enough" true
```

or
 ```
qc::response record invalid password "" "Password not complex enough"
```

**Comment by bvw** [comment]
Another way of extending the API could be 
`qc::response record sensitive password`

`qc::response record invalid password secret "Password not complex enough" true`

**Comment by daniel-qcode** [comment]
```
# Current
qc::response record invalid $name $value $message

# Option 1 - Would not be backwards compatible
qc::response record invalid name $name value $value message $message
qc::response record invalid name $name sensitive $value message $message

# Option 2 - Could potentially be backwards compatible
qc::response record invalid $name $value $message
qc::response record invalid -sensitive $name $value $message

#Once someone flags a field as sensitive should they be able to remove this flag by subsequent calls:
qc::response record valid -sensitive password mypassword "Password Valid"
qc::response record valid password mypassword "Password Not Complex Enough"
```

daniel-qcode:
Updated

PR #325, opened 2016-01-27
Support qc::upset $level $varname UNDEF

PeterChaplin:


PeterChaplin:
Updated with automated tests

PR #333, opened 2016-02-11
Try to ensure we're testing the correct version

davidqc:


davidqc:
Note:
```package ifneeded qcode [package present qcode]``` will return a list of the source files used to load that version.
eg.
```
source /usr/lib/tcltk/qcode6.41.0/dict.tcl
source /usr/lib/tcltk/qcode6.41.0/filter.tcl
....
```
or
```
source ./package/dict.tcl
source ./package/filter.tcl
....
```

PR #346, opened 2016-03-03
Added url_path into Tcl ensemble

priyankjalu89:
- [x] Security: Is user input sanitised ?
- [x] Security: Is substitution protected from injection attacks ?

### DIFF
--- a/tcl/cast.tcl
+++ b/tcl/cast.tcl
@@ -695,4 +695,17 @@ namespace eval qc::cast {
         }  
         return -code error -errorcode CAST "Could not cast $string to an url."      
     }
+    
+    proc url_path {string} {
+	#| Cast the given string to an url_path
+	#| (See also qc::is url_path)
+	set lower [string tolower $string]
+	if { [qc::is url_path $lower] } {
+	    return $lower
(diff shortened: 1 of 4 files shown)


tcl/cast.tcl
```diff
@@ -695,4 +695,17 @@ namespace eval qc::cast {
         }  
         return -code error -errorcode CAST "Could not cast $string to an url."      
     }
+    
+    proc url_path {string} {
+	#| Cast the given string to an url_path
+	#| (See also qc::is url_path)
+	set lower [string tolower $string]
```
nicky-johnstone: I've noticed that `qc::cast url` and `qc::cast relative_url` and that's why the same is done here but we might want to consider removing that behaviour for all URI casting because URIs can [contain some case-sensitive elements](https://tools.ietf.org/html/rfc3986#section-6.2.2.1) (the path typically is case-sensitive but this also depends on the server/file-system capabilities).

E.g. `/Hello/World/` is not necessarily the same as `/hello/world/`

I'm not sure if this is appropriate to change now or if a new ticket should be created to consider this change.

PR #352, opened 2016-03-30
Add -not option to db_row_exists, and add sql_where_not proc

PeterChaplin:


PeterChaplin:
For example:
```tcl
# Validating input for product update
if { [db_row_exists -not {~ product_id} product ~ product_code] } {
  error "Another product has code $product_code" {} USER
}
```

PeterChaplin:
Updated with tests (passed on DEV)

nicky-johnstone:
What would the benefits of the `-not` flag be versus something like:


```tcl
if { ! [db_row_exists product ~ product_id product_code]
       && [db_row_exists product ~ product_code] } {
    ...
}
```

The only benefit that I can think of is that it compacts it down to one line and one query instead of two. Is that enough of an advantage to warrant the addition of the `-not` flag?

PeterChaplin:
For checking uniqueness constraints, I don't think there are any other differences.

I can imagine other situations where we might want to use a `-not` flag, but I can't recall any of them actually occurring, so it might be better to leave this for now.

daniel-qcode:
Unnecessary for now ... closing

PR #370, opened 2016-05-03
Log invalid form variable names

nicky-johnstone:
**Tested on mla-erp**

Input:
```
https://qcode.dev.mlaccessories.co.uk/home.html?::testing=123
```

Log:
```
[03/May/2016:09:58:09][4005.7f2bd0507700][-conn:mla_erp:3-] App:Notice: Invalid form variable name: ::testing
```

Input:
```
https://qcode.dev.mlaccessories.co.uk/home.html?::testing=123&foo::bar=baz
```

Log:
```
[03/May/2016:10:00:57][4005.7f2bd0507700][-conn:mla_erp:3-] App:Notice: Invalid form variable name: ::testing
[03/May/2016:10:00:57][4005.7f2bd0507700][-conn:mla_erp:3-] App:Notice: Invalid form variable name: foo::bar
```

### DIFF
--- a/tcl/filter.tcl
+++ b/tcl/filter.tcl
@@ -161,7 +161,7 @@ proc qc::filter_authenticate {event args} {
 }
 
 proc qc::filter_http_request_validate {event {error_handler "qc::error_handler"}} {
-    #| Check that request string and connection url are both valid.
+    #| Check that request string, connection url, and form variable names are valid.
     qc::setif error_handler "" "qc::error_handler"
     ::try {
         set request [ns_conn request]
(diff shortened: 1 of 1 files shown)


tcl/filter.tcl
```diff
@@ -174,6 +174,14 @@ proc qc::filter_http_request_validate {event {error_handler "qc::error_handler"}
             return [ns_returnbadrequest "\"$url\" is not a valid URL."]
             return filter_return
         }
+
+        set names [qc::form_var_names]
+        foreach name $names {
+            if { [regexp {::} $name] } {
+                log "Invalid form variable name: $name"
```
nicky-johnstone: Simply logging the name only. Are we only going to be using this to measure frequency of invalid form variable names or do we wish to log more detail for further investigation?

daniel-qcode: Can we log $url too please ?

tcl/filter.tcl
```diff
@@ -174,6 +174,14 @@ proc qc::filter_http_request_validate {event {error_handler "qc::error_handler"}
             return [ns_returnbadrequest "\"$url\" is not a valid URL."]
             return filter_return
         }
+
+        set names [qc::form_var_names]
+        foreach name $names {
```
daniel-qcode: Can we add a comment to explain why we are checking if name contains ::

nicky-johnstone:
Updated. Logged the full request rather than just the URL as the request can tell us the method and full query string too.

PR #389, opened 2016-06-21
Response redirect

daniel-qcode:
NOT Ready for merge ... completely untested.


daniel-qcode:
@bvw was thinking about something like this to protect against external redirects.

qc::return_next seems to convert relative URLs to absolute URLs if the Host and Post are known ... do we want the same behavior here ... I can't think of a reason why we do this for qc::return_next can you ?

daniel-qcode:
Ready for merge

PR #403, opened 2016-07-19
Improve sql-injection protection for sql_sort

PeterChaplin:
- Tests run sucessfully
- Tested a report with sql_sort on mla_erp DEV
- Audited mla-erp code for unusual usage of sql_sort, nothing found.

### DIFF
--- a/tcl/sql.tcl
+++ b/tcl/sql.tcl
@@ -63,6 +63,7 @@ proc qc::sql_sort { args } {
     set order_by_list {}
     for {set i 0} {$i<[llength $list]} {incr i} {
 	set this_item [lindex $list $i]
+        set this_item [qc::db_quote_identifier $this_item]
 	set next_item [lindex $list [expr {$i+1}]]
 	switch -nocase $next_item {
 	    ASC {
@@ -129,7 +130,7 @@ proc qc::sql_sort { args } {
         }
         upset 1 offset $offset
 
-        return "$sql limit $limit offset $offset"
+        return "$sql limit [qc::db_quote $limit] offset [qc::db_quote $offset]"
     } else {
 	return $sql
     }
--- a/test/sql.test
+++ b/test/sql.test
@@ -41,17 +41,27 @@ namespace eval ::qcode::test {
     test sql_sort-1.0 {sql_sort} -setup {
     } -body {
         sql_sort name email user_code 
-    } -cleanup {} -result {name,email,user_code}
+    } -cleanup {} -result {"name","email","user_code"}
 
     test sql_sort-1.1 {sql_sort with order} -setup {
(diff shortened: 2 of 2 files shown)


test/sql.test
```diff
@@ -41,17 +41,27 @@ namespace eval ::qcode::test {
     test sql_sort-1.0 {sql_sort} -setup {
     } -body {
         sql_sort name email user_code 
-    } -cleanup {} -result {name,email,user_code}
+    } -cleanup {} -result {"name","email","user_code"}
```
daniel-qcode: what happens when you call something like `qc::sql_sort column_a DESC NULLS LAST`?
Is this pattern ever used for MLA?

daniel-qcode: Guess this doesn't apply cause you would just use the `-nulls` option for this.
Can you just quickly check MLA to make sure though?

tcl/sql.tcl
```diff
@@ -129,7 +130,7 @@ proc qc::sql_sort { args } {
         }
         upset 1 offset $offset
 
-        return "$sql limit $limit offset $offset"
+        return "$sql limit [qc::db_quote $limit] offset [qc::db_quote $offset]"
```
daniel-qcode: should we set the type for integer when calling db_quote for limit and offset?

test/sql.test
```diff
@@ -41,17 +41,27 @@ namespace eval ::qcode::test {
...
 
     test sql_sort-1.1 {sql_sort with order} -setup {
     } -body {
         sql_sort name DESC email user_code 
-    } -cleanup {} -result {name DESC NULLS LAST,email,user_code}
+    } -cleanup {} -result {"name" DESC NULLS LAST,"email","user_code"}
 
     test sql_sort-1.2 {sql_sort with paging} -setup {
     } -body {
```
daniel-qcode: can you add a test case for the `-nulls` option.

test/sql.test
PeterChaplin: I'm pretty sure that would have been an error under the existing code, this code shouldn't affect that.

The correct usage would be `qc::sql_sort -nulls last -- column_a DESC`, which should continue to work just fine (see test case 1.1)

PeterChaplin: The existing code would generate:
`column_a DESC, NULLS, LAST` which would be a sql error (note the commas)

tcl/sql.tcl
PeterChaplin: I don't think it's necessary.

daniel-qcode: ok

test/sql.test
PeterChaplin: Added

PR #405, opened 2016-07-26
Enable thread-cache for perm_test_user

PeterChaplin:


PeterChaplin:
@bvw Did we reach a conclusion on this?

bvw:
Is the profile of normal usage cases that a thread would make one or a small number of unique calls to perm_test ?
If that is the case then the overhead of caching is actually detrimental in most cases.

I think you said that perm tests inside loops could be refactored to improve performance and my inclination is to go in that direction.

However, if the profile of requests to perm over the normal lifetime of a thread makes many repeated identical calls then that would make a strong case for accepting this change.

PeterChaplin:
I don't really have an overview of `perm_test_user` within any of our projects, I'd have to add some sort of logging to measure that. I can say that this change has been the biggest performance improvement to mla-erp `/product.html` that I've managed to implement on DEV, and `/product.html` has the highest total runtime on mla-erp pretty consistently.

In theory, any perm tests inside iteration loops could be refactored, I'm just concerned that doing so would involve a lot of code re-factoring, which is going to take more time, and introduce more complexity and opportunities for errors.

Another approach would be to make this optional, eg. `perm_test_user -cache $user_id $perm_name $method`. There'd still be some extra work needed to propagate that change to other procs (such as `perm_if`), but I think it would make re-factoring iteration and frequently-called procs easier.

PR #406, opened 2016-08-02
Trim sql_sort args before quoting

PeterChaplin:


daniel-qcode:
What issue does this fix?

PeterChaplin:
Encountered problems on DEV:
https://github.com/qcode-software/mla-erp/issues/1822
I don't think the sql_injection work has been rolled out to LIVE yet, so it's not currently operational.
The existing sql_sort code does make an effort to support comma-separated arguments, though,

PeterChaplin:
Test case added

daniel-qcode:
@PeterChaplin Have you tested the case where the identifier has a alias eg `sql_order p.product_code`?

If this doesn't work then please raise GitHub issue to address.

PeterChaplin:
Issue raised: https://github.com/qcode-software/qcode-tcl/issues/407

PR #415, opened 2016-08-15
Have lequal check that args can be parsed to lists before doing so

PeterChaplin:


test/util_list.test
```diff
@@ -213,6 +213,32 @@ namespace eval ::qcode::test {
...
+    test lequal-1.2 {lequal different lengths} -setup {
+    } -body {
+        lequal [list a b c ""] [list a b c]
+    } -cleanup {} -result {0}
+
+    test lequal-1.3 {lequal nested} -setup {
+    } -body {
+        lequal [list [list a b] [list c d]] [list [list a b] [list c d]]
+    } -cleanup {} -result {1}
+
```
daniel-qcode: can you add a case for nested that doesn't match ?

tcl/util_list.tcl
```diff
@@ -103,9 +103,22 @@ proc qc::lunique {list} {
 proc qc::lequal {a b} {
     #| Compare 2 list 
     # Author RS Tcl wiki
-    if {[llength $a] != [llength $b]} {return 0}
-    if {[lindex $a 0] == $a} {return [string equal $a $b]}
-    foreach i $a j $b {if {![lequal $i $j]} {return 0}}
+    if { ! [string is list $a]
```
daniel-qcode: can you add a short comment to each if statement eg. `#a or b not a list`, `a and b not the same length` etc

PeterChaplin:
Updated.

PR #418, opened 2016-08-25
Content disposition bug fix

PeterChaplin:


tcl/return.tcl
```diff
@@ -50,10 +50,10 @@ proc qc::return2client { args } {
     # Content-Disposition
     if { $download } {
         # Download payload as an attachment
-        default content-disposition "attachment; filename=$filename"
+        default content-disposition "attachment; \"filename=$filename\""
```
bvw: What does this structured string say we should do if $filename contains a quote ?

PeterChaplin: http://stackoverflow.com/questions/93551/how-to-encode-the-filename-parameter-of-content-disposition-header-in-http

From the looks of it, that depends on which browser we are returning to.

I suspect that the best solution would simply be to white-list allowed filename characters, and strip out the rest.

bvw: Okay, I'd prefer to raise an error if the $filename was problematic and enforce this restriction in the application when files are being uploaded.
Can we enforce a filename restriction in the data model ?

PR #419, opened 2016-09-02
Csv processing

davidqc:


bvw:
Does this also need back ported to qcode-tcl-5 ?

davidqc:
No, not for now

PR #431, opened 2016-12-07
Add qc::response data proc and test file

PeterChaplin:


tcl/response.tcl
```diff
@@ -1,18 +1,30 @@
...
 
-    namespace export status record message action extend
+    namespace export status record message action extend data
     namespace ensemble create -unknown {
         response_subcommand_map
     }
 
+    proc data {tson} {
+        #| Sets the JSON response object data property using tson
+        global data
```
bvw: Is this global the one Nicky already chose to write the response object to or a new global ?

tcl/response.tcl
```diff
@@ -1,18 +1,30 @@
...
+
     proc extend {name args} {
         #| Extends the JSON response with an object named $name with properties defined in $args.
         if { [llength $args] % 2 != 0 } {
             return -code error "Usage: qc::response extend name key value ?key value ...?"
         }
         global data
-        dict for {key value} $args {
-            dict set data $name $key $value
+        if { $name eq "data" } {
```
bvw: Do we need this feature ? is it another funky way to call the data method ? Does it prevent a user using "data" key ?

tcl/response.tcl
```diff
@@ -383,6 +400,9 @@ proc qc::response2html_snippet {} {
                 status {
                     set status $value
                 }
+                data {
+                    set data_html [qc::response_tson2html_fragment $value]
```
bvw: Help me understand data_html please.

tcl/response.tcl
PeterChaplin: Backwards-compatibility for anyone using `qc::response extend data foo bar`, will result in the same functionality.

tcl/response.tcl
PeterChaplin: It is the HTML fragment which represents the `data` property.

PeterChaplin: `response2html` needs to handle the `data` property somehow (even if just by ignoring it), I went with including a `<div>` with class `data`, containing:

1. Each property of an object as a div, with an id for the property name and the div contents for the value
2. Each element of an array as a div with class "array-element", the contents being the element value
3. Anything explicitly declared as a string, number, or boolean, wrapped in a span with a class to indicate the data type
4. Anything else included directly, as escaped html.

PeterChaplin:
The documentation will need updating, do you want to put this request on hold until I have one for the docs as well?

bvw:
Would be nice to get this page updated for example and any others that are relevant.
https://github.com/qcode-software/qcode-tcl/blob/master/doc/response_api.md

PeterChaplin:
Updated with documentation.

tcl/response.tcl
daniel-qcode: I think if we wanted to set the response data using tson we would just use `qc::response data $data` .
If the developer used `qc::response extend data value` then I think it is reasonable that we keep the existing behavior that does not support tson structures.

I'd imagine that we will most probably use `qc::response data ...` instead of `qc::response extend ...` going forward.

Any other situations that you think this functionality would be required.

PeterChaplin: This code change is required in order to keep the existing behaviour, in case any code currently uses `qc::response extend data name value` (which is the existing usage).

I would expect this usage to be deprecated in favour of `qc::response extend ..." going forward.

tcl/response.tcl
```diff
@@ -405,11 +425,50 @@ proc qc::response2html_snippet {} {
                            [h div class message [join $message_elements \n]] \
                            [h div class record [join $record_elements \n]] \
                            [h div class action [join $action_elements \n]] \
+                           [h div class data $data_html] \
                            [h div class extended [join $extended_elements \n]]
                           ] \n] \
                ]
 }
 
+proc qc::response_tson2html_fragment {tson} {
```
PeterChaplin: I have no strong opinion; Any arguments in favour of one name over another?

PeterChaplin: It's not returning a full HTML document, and the tson being passed in isn't necessarily the full data.data property.

tcl/response.tcl
daniel-qcode: As far as I can tell qc::response extend is only used in one place in tlc-www ... can you double check mla-www and mla-erp ... we might be able to deprecate qc::response extend to simplify the code and bump the qcode-tcl version to indicate a breaking change

test/response.test
```diff
@@ -90,15 +90,15 @@ namespace eval ::qcode::test {
 <div class="message"></div>
 <div class="record"></div>
 <div class="action"></div>
```
daniel-qcode: Could we maybe have something like this for object-property elements ... will call you in a bit to discuss:

```
<div class="object-property" data-name="surname" data-value="O'Neil" data-type="string"</div>
```

daniel-qcode: Hmm guess that wouldn't work if the property's value is an array or an object ... will call to discuss in a bit anyway

doc/connection-response.md
```diff
@@ -59,7 +63,7 @@ Details on setting up the response can be found in the [API].
 Extending The Response
 ----------------------
```
daniel-qcode: Can we add a deprecated comment here too please

doc/procs/response_data.md
```diff
@@ -0,0 +1,42 @@
+qc::response data
+====
+
+part of [Connection Response API](../response_api.md)
+
+Usage
+-----
+`qc::reponse data data_tson`
```
daniel-qcode: Is this missing a $ symbol: `qc::reponse data $data_tson`

PR #432, opened 2016-12-16
Relax criteria for external domain redirect

priyankjalu89:
The security issue with redirects is for open redirects based on user-supplied urls.

This change relaxes the constraint on redirect destinations but relies on the caller to sanitise the url destination.

https://portswigger.net/KnowledgeBase/issues/Details/00500100_Openredirection

### DIFF
--- a/tcl/response.tcl
+++ b/tcl/response.tcl
@@ -144,7 +144,7 @@ namespace eval qc::response {
     ##################################################
     namespace eval action {
         
-        namespace export redirect resubmit login
+        namespace export redirect external_redirect resubmit login
         namespace ensemble create
 
         proc redirect {url} {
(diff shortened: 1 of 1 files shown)


priyankjalu89:
Built local version of package and tested redirect.

daniel-qcode:
I'd like to keep this check if possible as it mirrors the same protection built into qc::return_next.

Would it be possible to add an extra method to qc::response instead something like "qc::response external_redirect" instead.

Another alternative to fix just this case would be to extend the response using `qc::response data external_redirect` and handle the redirect ourselves in Javascript by listening for the ValidationComplete event and inspecting the data in the response.

On the existing checkout we can't use qc::return_next for the same reason and instead have to use a lower level command called "ns_returnredirect"

PR #435, opened 2017-01-10
A postgresql spelling corrector based on levenshtein distance

davidqc:


tcl/util_string.tcl
```diff
@@ -229,3 +229,66 @@ proc qc::strip_common_leading_whitespace {text} {
...
+    #|         create table spell_corpus (
+    #|                  word text,
+    #|                  entries int
+    #|                  );
+
+    # insert into temp table since ts_stat takes a sql query as parameter
+    db_dml {drop table if exists temp_corpus}
+    db_dml {create table temp_corpus(words text)}
+
+    set fh [open $corpus_file r]
```
bvw: I'd prefer to pass in the corpus text and push responsibility for file read to the caller.

tcl/util_string.tcl
```diff
@@ -229,3 +229,66 @@ proc qc::strip_common_leading_whitespace {text} {
...
+        where 
+        levenshtein(lower(word),lower(:word))<=2
+        order by levenshtein(lower(word),lower(:word)) asc,entries desc 
+        limit 1
+    }
+    db_0or1row $qry {
+        set suggestion ""
+    } 
+
+    if { $suggestion ne $word } {
```
bvw: If the word was correctly spelt what would be returned ?

davidqc: It only generates output if a correction has occurred.

It looks for the most frequently occurring word with the shortest distance within a levenshtein distance of 2. If that word is the same as the input word, it returns an empty string - this is analogous to the behaviour of Norvig version we use.

bvw: ok

davidqc:
updated

PR #449, opened 2017-03-22
Check if resize is needed before performing

PeterChaplin:


tcl/image_file.tcl
```diff
@@ -89,12 +89,17 @@ proc qc::image_resize {args} {
         file delete $tmp_file
     }
 
-    set tmp_file $file
-    set file [qc::image_file_resize $tmp_file $max_width $max_height]
-    file delete $tmp_file
-
     dict2vars [qc::image_file_info $file] width height
 
+    if { $width > $max_width
```
bvw: Please add a code comment to explain

PR #451, opened 2017-03-22
Faster db_file_export proc

PeterChaplin:


bvw:
I'm happy to make this technique an alternative but not to replace the current simpler technique.

PR #462, opened 2017-06-14
Remove hard-coded database from db_pg_copy_load

PeterChaplin:
Any code currently using db_pg_copy_load will need a re-write to use this.


daniel-qcode:
@PeterChaplin please create a breaking release

PR #484, opened 2017-09-21
Excel images

PeterChaplin:


PeterChaplin:
Tested with local make-install by adding an image to an excel file, including test with/without x/y pair and x_scale/y_scale pair.

tcl/excel.tcl
```diff
@@ -204,7 +282,7 @@ proc qc::excel_file_create {args} {
     ::try {
         log Debug $script_filename
         exec_proxy -timeout $timeout perl $script_filename
-        file delete $script_filename
+        #file delete $script_filename
```
nicky-johnstone: Is this meant to still be commented out?

PeterChaplin: No, fixed.

PR #489, opened 2017-09-28
db_col_varchar_length multi-schema support

PeterChaplin:


bvw:
Can we support usage that optionally allows tables to be qualified with a schema which may not be the default ?
`qc::db_col_varchar_length users firstname`
`qc::db_col_varchar_length web.users firstname`
or
`qc::db_col_varchar_length web users firstname`

PeterChaplin:
That would be possible, but at this stage it's a feature that we don't need. Could we leave that for a later version? I feel like if we're going to support it here, we should go through all the introspection procs where a table name is passed in and add support, which could add a fair bit of complexity and development time.

bvw:
I'd like to compare 2 versions for one proc to see how they compare. 
Supporting the schema as an argument with a default may be very simple especially if we can abstract out an introspection proc to return the current schema.
The sql in this pull request takes some head scratching to make sense of and I'm still not clear about how it works.

PR #511, opened 2017-11-23
Case insensitive headers

davidqc:


daniel-qcode:
will probably need the same changes in latest qcode-tcl too

PR #514, opened 2017-12-01
postcode_parse

davidqc:


davidqc:
Update

tcl/sql.tcl
```diff
@@ -181,6 +181,31 @@ proc qc::sql_list2array { args } {
     }
 }
 
+proc qc::postcode_parse { postcode } {
```
daniel-qcode: could we move into util.tcl or new file

davidqc:
update

PR #523, opened 2018-03-20
Backport tson from master

alinoil:


tcl/tson.tcl
```diff
@@ -1,5 +1,54 @@
...
+    }
+    return [list number $value]
+}
+
+proc qc::tson_boolean {value} {
+    #| Returns a TSON boolean.
+    if { [string tolower $value] in [list "" "null"] } {
+        return "null"
+    }
+    return [list boolean [qc::cast boolean $value true false]]
```
daniel-qcode: `qc::cast boolean` isn't supported in qcode-tcl-5 ... think it needs to be `qc::cast_boolean` (with underscore)

alinoil: Pushed it with a cast_boolean

PR #533, opened 2018-08-23
Updating the form to use the authenticity token

adam-bethell:
Added a section on updating the form from Tutorial 2. 
Also added a couple of links, sentences for clarification.


adam-bethell:
Updated form example to use qc::form

adam-bethell:
I've taken out the use of `qc::h` so the change from tutorial 2 should be much smaller now.

PR #579, opened 2019-01-10
Support interval type for is, cast, and castable

PeterChaplin:


bvw:
Thanks Peter, could you please add some tests in the test folder.

PeterChaplin:
@bvw tests added

PR #586, opened 2019-02-07
Cast/Validate Bug Fix for qcode-tcl data types

daniel-qcode:
Always qcode-tcl data types using qcode-tcl validate/cast procs even when prefixed with db schema.

Ready for merge


daniel-qcode:
@bvw  same change for qcode-tcl version 8

PR #595, opened 2019-05-09
tabular_text_parse trim option - qcode-tcl-5

adam-bethell:
Adds `trim false` and `true` to columns conf.
E.G.
```tcl
lappend columns_conf [list label "Column A" var_name col_a trim false]
```


daniel-qcode:
can you add extra tcltests for tabular_text_parse to cover this new option usage

PR #598, opened 2019-05-29
backporting changes to qcode-tcl

davidqc:
backport latest changes to qcode-tcl-5


daniel-qcode:
@davidqc should we roll this out to tlc-erp and tlc-lib next ... considering regexp_escape change?

PR #600, opened 2019-05-31
Hba sessions

davidqc:


daniel-qcode:
@davidqc can we test this together, I'm not sure how this will work when the "hba session" expires, is it possible for user to recover ... from reading the code I think we will detect the session cookie and always attempt session authentication instead of falling back to hba, so when session expires "hba" users wouldn't be issued a new session automatically.

davidqc:
It's not particularly clean testing that scenario.

First we get the expected expiration screen:

image

Then an error where we don't seem to know the session is expired:

image

Then on refresh, it recovers:

image

So definite room for improvement there

daniel-qcode:
@davidqc will need pull request to `/logout.tcl` bugfix in tlc-erp too.

PR #619, opened 2019-08-29
add restful blog tutorial and link to it in README

connelallison:


doc/tutorial-restful-blog.md
```diff
@@ -0,0 +1,314 @@
...
+}
+```
+
+After saving and restarting, try submitting a blog post. You should, at last, be able to see the post you have submitted (and you should be able to see your previous posts if you change your address to have their entry_id instead). You have successfully added basic create and read functionality to your site. 
+
+## register, h, and form
+
+Before moving on, let's review some of the procs we have been using in our code.
+
+We have made use of the `register` proc throughout the `url_handlers.tcl` file. This is used to register a path, so that the server has instructions in place for how to deal with, for example, a request such as `GET entries/1`. Note the used of colon variables, for example in your handler for viewing an entry:
```
bvw: Can we link to documentation on the `register` proc ?

bvw: Sorry it is linked below

connelallison: I link here:

https://github.com/qcode-software/qcode-tcl/blob/blog_tutorial/doc/registration.md

at the end of the discussion of the register proc - is there other documentation I should also be linking? I didn't see a register.md in the procs folder.

doc/tutorial-restful-blog.md
```diff
@@ -0,0 +1,314 @@
...
+
+```tcl
+register GET /entries/:entry_id {entry_id} {
+    #| View an entry
+    return [entry_get $entry_id]
+}
+```
+
+By putting a colon at the start of `:entry_id`, we instruct the handler to treat it as a variable name and pass its value to `entry_id` to be used in the function body. For more detail on the register proc, [see its documentation](registration.md).
+
```
bvw: I suggest: For more detail, see [documentation on the register proc](registration.md)

It is preferable that  hyperlink context provided stands on its own.

"see its documentation" requires the context of the sentence it is in.

connelallison: Alright - I'll make a similar change for the other links

doc/tutorial-restful-blog.md
```diff
@@ -0,0 +1,314 @@
...
+When constructing our form (and later in the `entry_get` proc), we used the `h` proc to generate html elements for us. The first argument you pass it is the type of html element you want. After specifying the type, any additional elements will be interpreted as alternating key value pairs. If the final argument is unpaired, it is placed in the body of the element. Consider this example:
+
+```tcl
+h a href "http://localhost/entries/new" "Submit another blog"
+```
+
+This will return a string containing the HTML for an `<a>` element that reads "Submit another blog" and links back to the new entry form. It is preferable to use the `h` proc instead of writing strings of raw HTML yourself, and essential for anything that involves variable substitution - aside from making construction of HTML easier, it also takes care of sanitising your input data and preventing critical security vulnerabilities such as [injection attacks](https://en.wikipedia.org/wiki/Code_injection#Examples). Additional examples can be found in [its documentation](procs/h.md).
```
bvw: Additional examples can be found in the [qc::h documentation](procs/h.md)

doc/tutorial-restful-blog.md
```diff
@@ -0,0 +1,314 @@
...
+
+We also used the `form` proc to construct the form you passed to the user. It will return a form element, using the arguments you pass it as key-value pairs and the final unpaired argument you pass it (if applicable) placed in the body of the element. See our example:
+
+```tcl
+return [qc::form method POST action /entries $form]
+```
+
+Here, it returns a `<form>` element with `method="POST"` and `action="/entries"`, and then places the html we have stored in the `form` string variable inside the body of the form. Use this proc when you are constructing forms - aside from not having to write out the full HTML, it also takes care of attaching a hidden authenticity token - if you see an error that refers to there being no authenticity token, you should check to see if you have skipped over using this proc (or a similar one), which would have taken care of it for you. See its [full documentation](procs/form.md).
```
bvw: Same for link to form doc.

doc/tutorial-restful-blog.md
```diff
@@ -0,0 +1,314 @@
...
+The handler knows to check for an input named `_method` when it receives the POST request, and after finding that `_method` exists and has a value of PUT, treats the POST request as a PUT request.
+
+Similarly, when we use the `form` proc to create our "Delete" button, we use a hidden `_method` input - in this case however, we let the `form` proc handle this for us by simply including "method DELETE" in the `form` proc's arguments:
+
+```tcl
+    append html [form method DELETE action /entries/$entry_id \
+		     [h input type submit name submit value "Delete this blog"]]
+```
+
+If you refer back to the `form` proc's [documentation](procs/form.md), you will see in the second example how the proc inserts the hidden input for us (in addition to the authenticity token).
```
bvw: [`form` proc's documentation](procs/form.md)

PR #626, opened 2019-11-28
email2multimap quoted string handling

PeterChaplin:


tcl/util.tcl
```diff
@@ -708,4 +750,133 @@ proc qc::regexp_escape {string} {
...
+        }
+    }
+    return $list
+}
+
+proc qc::split_pair2 {args} {
+    #| Split a string into 2 parts at the first occurence of the delimiter,
+    #| ignoring quoted or escaped delimiters
+    qc::args $args -nocomplain -escape \\ -quote \" -- string delimiter
+    default nocomplain false
```
bvw: Rather than complain would it be possible to test the condition of the string first to anticipate if it can be split or not.

PeterChaplin: It would be possible, but the code for testing would have roughly the same complexity as the code for splitting, and there would be a lot of duplication between the two. Overall I'd say it would increase the complexity of the code we'd need to maintain going forward.

tcl/util.tcl
```diff
@@ -708,4 +750,133 @@ proc qc::regexp_escape {string} {
     } 
 
     return [string map $list $string]
+ }
+
+proc qc::split2 {args} {
```
bvw: Does this function essentially convert a string to a list ?

PeterChaplin: Yes

bvw: Does this do the same thing:-
```Tcl
proc foo {string} {
set result [list]
foreach part $string { lappend result $part }
return $result
}
```

PeterChaplin: Not exactly, this proc takes the delimiter to split on as an argument, the code you've given would use any whitespace as the delimiter.

bvw:
Can you give an example of usage you need please.

PeterChaplin:
Email file "1574779303.Vca02I675efM228364.dreadnought" on dreadnought failed to insert with error `Delimiter ":" was not found in the string "\"`, this was due to an email header:
```
Content-Type: application/octet-stream;
 name="=?utf-8?B?PT91dGYtOD9CP1F6cGNWWE5sY25OY2QybHVOeTB4WEVSbGMydDBiM0Jj?=\
\

 =?utf-8?B?NllLdTU2NnhMUzNsaTcvbGlLQmNNakF4Pz0NCiA9P3V0Zi04P0I/T1RF?=\
\

 =?utf-8?B?eE1EY3RMZVM0dSttaW1GeFJVZWFJcXVXYnZqSXdNVGt4TVRBM01UUXhN?=\
\
 =?utf-8?B?ek13TG5CdVp3PT0/PQ==?="
```
Looking at the RFC, I believe newlines within quotes or after `\`s should not be used for splitting header lines.

This pull request is to change `qc::email2multimap` to account for quotes and escapes when separating the email head from body, and when splitting the head into separate lines.

For this, I have added the procs `qc::split_pair2` and `qc::split2`, which function the same as `qc::split_pair` and `split`, but ignoring delimiters within quotes and after escape characters.

PeterChaplin:
@bvw updated

PR #629, opened 2020-03-02
Improve qc::is date

PeterChaplin:
2019-11-31 is not a valid date, but we do cast it to 2019-12-01
Tcl accepts it as a date, postgresql does not.


PeterChaplin:
@bvw Updated with test for leap year

bvw:
Thanks, are all tests passing ? for
Check leap year
Check is_date

PeterChaplin:
I ran `make test VERSION=10.9.1` and it all passed ok.

PR #653, opened 2020-10-14
Allow schema option on qc::perm_set

PeterChaplin:


bvw:
What was the error that made you look at this ?

PR #654, opened 2020-10-14
Add qc::dict_mappings_equal and qc::ldict_mappings_equal

PeterChaplin:


bvw:
Please add tests under test/dict.test

PeterChaplin:
Tests added and run ok on jessie

bvw:
I think you have chosen these proc names because the dicts are not identical.
I think equality is different from being identical.
dicts_equal 
dicts_identical

PeterChaplin:
My concern is that a dict is an ordered mapping, the standard definition of dict equivalence would be the same mappings in the same order.
For comparison, the `lequal` proc checks that two lists have the same elements in the same order. In cases where order is irrelevant, we use `lsort` first. We do also have a `qc::dict_sort` proc, so we could use the same approach.
Re-defining dict equality to only consider mappings and not ordering is an option, but could be the source of some confusion.

bvw:
There are other examples of where equality is not `===` checking if objects are identical.
`2==2`
`2.0==2`
`"2.0"==2`
`Number("2.0")==2`

PeterChaplin:
`lequal {1 2 3} {3 2 1}` returns false.

bvw:
We could make casting explicit for example
`string equal FOO foo`
`string equal [upper FOO] [upper foo]`

`qc::dicts_equal [qc::dict_sort $dict1] [qc::dict_sort $dict2]`

PeterChaplin:
@bvw I have updated the pull request to add `qc::dicts_equal` as a proc

tcl/dict.tcl
```diff
@@ -135,4 +135,24 @@ proc qc::dict_zipper {keys values} {
...
+        if { $key1 ne $key2
+             ||
+             $value1 ne $value2 } {
+            return false
+        }
+    }
+    return true
+}
+
+proc qc::dict_mappings_equal {dict1 dict2} {
```
bvw, line 155: I think we can remove this proc and ask the caller to implement the line below.

tcl/ldict.tcl
```diff
@@ -105,3 +105,25 @@ proc qc::ldict2tbody {ldict colnames} {
     return $tbody
 }
 
+proc qc::ldict_mappings_equal {ldict1 ldict2} {
```
bvw, line 108: What would qc::ldict_equal look like ?

PeterChaplin: It would look the same, but with `qc::dicts_equal` instead of `qc::dict_mappings_equal`.
It would be hard for the caller to specify that the dicts need to be sorted before comparison.

PR #674, opened 2021-01-05
3883 remittance parser errors

PeterChaplin:


bvw:
Is there an S3 test suite ?

PeterChaplin:
@bvw yes, we have a test suite

bvw:
Can you paste in the results of the tests passing here and I can then merge.

PeterChaplin:
Do you have an example of what the `~/.qcode-tcl` file should look like and which s3 test bucket I should be using?

PeterChaplin:
image

PR #692, opened 2021-05-06
Add error checking when using pg_tcl

PeterChaplin:


PeterChaplin:
Tests:
image

Change in behaviour - `qc::db_select_table` will throw a tcl error if the sql request errored while using `db_connect` (currently, `pg_exec` does not error if the query fails, but only returns no results)

PR #700, opened 2021-07-13
fix tcltest

daniel-qcode:
* preserve qc::ldelete behaviour of preventing errors when attempting to delete a index that is greater than the max list index
* update relative file paths to be anchored by user's home directory
* drop test db user if exists before attempting to create it


daniel-qcode:
READY for merge

image

PR #704, opened 2021-07-29
Added Proc To Convert HTML To SVG

nicky-johnstone:
Testing
---
```
tclsh html.test
html.test:	Total	40	Passed	40	Skipped	0	Failed	0
```


bvw:
Doesn't pass barrier for inclusion in library.

PR #715, opened 2022-02-03
New type next_url

priyankjalu89:
image

### DIFF
--- a/tcl/cast.tcl
+++ b/tcl/cast.tcl
@@ -870,7 +870,7 @@ namespace eval qc::cast {
         
         # check that redirection is to the same domain
         set conn_host [qc::conn_host]
-        if { ![regexp "^https?://${conn_host}(:\[0-9\]+)?(/|\$)" $next_url] } {
+        if { ![regexp "^https?://${conn_host}(:\[0-9\]+)?(\\?|/|\$)" $next_url] } {
             return -code error -errorcode CAST "Could not cast \"[html_escape $next_url]\" to valid next_url. Domain must be \"$conn_host\"."
         }       
 
--- a/tcl/cast.tcl
+++ b/tcl/cast.tcl
@@ -281,7 +281,7 @@ proc qc::data_type_error_check {data_type value} {
 
 namespace eval qc::cast {
     
-    namespace export integer bigint smallint decimal boolean timestamp timestamptz char varchar text enumeration domain safe_html safe_markdown date postcode creditcard period epoch url url_relative url_path s3_uri time interval
+    namespace export integer bigint smallint decimal boolean timestamp timestamptz char varchar text enumeration domain safe_html safe_markdown date postcode creditcard period epoch url url_relative url_path s3_uri time interval next_url
     namespace ensemble create -unknown {
         data_type_parser
     }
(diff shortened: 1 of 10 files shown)


tcl/cast.tcl
```diff
@@ -853,4 +853,27 @@ namespace eval qc::cast {
...
+        }
+
+        # check for malicious mal-formed url
+        if { ![qc::is url $next_url] } {
+            return -code error -errorcode CAST "Could not cast \"[html_escape $next_url]\" to valid next_url. Invalid URL."
+        }
+        
+        # check that redirection is to the same domain
+        set conn_host [qc::conn_host]
+        if { ![regexp "^https?://${conn_host}(:\[0-9\]+)?(\\?|/|\$)" $next_url] } {
```
daniel-qcode, line 873: do we also need to handle https://www.abc.com#test

tcl/cast.tcl
```diff
@@ -853,4 +853,27 @@ namespace eval qc::cast {
...
+
+        # check for malicious mal-formed url
+        if { ![qc::is url $next_url] } {
+            return -code error -errorcode CAST "Could not cast \"[html_escape $next_url]\" to valid next_url. Invalid URL."
+        }
+        
+        # check that redirection is to the same domain
+        set conn_host [qc::conn_host]
+        if { ![regexp "^https?://${conn_host}(:\[0-9\]+)?(\\?|#|/|\$)" $next_url] } {
+            return -code error -errorcode CAST "Could not cast \"[html_escape $next_url]\" to valid next_url. Domain must be \"$conn_host\"."
```
davidqc, line 874: $conn_host can be controlled by header manipulation so is not completely safe - does this end up being rendered back to the user? Is so, it need escaping (Maybe something further up takes care of that?)

priyankjalu89: Yes this end up being rendered back to the user and yes I think it should be the caller's responsibility to take care of that. Would it be okay to escape the next_url before it is passed to this cast api [here](https://github.com/qcode-software/qcode-tcl/pull/715/files#diff-bb1f05244dfece6f46f7e54ddff8d5a419586b13eca9f5294fc8a3229884a27fR135)

davidqc:
Can  we use ```qc::cast next_url``` within ```qc::response action redirect```?

tcl/cast.tcl
davidqc: We're pulling ```conn_host``` directly out of the headers ```[qc::conn_host]```within this proc so the caller can't deal with it.

priyankjalu89: Good point. I can `html_escape` the `string` at the start of the API if that helps ?

priyankjalu89: If we cannot trust the `conn_host` the only other option I can think of is using a `param`.

priyankjalu89:
> Can we use `qc::cast next_url` within `qc::response action redirect`? Unsure whether the errors presented to the user would become unsuitable.

Casting is something we do internally so providing and end user with a message e.g `could not cast <x> to Boolean` would not be a helpful message and it also confuse them as to what they should do next. `Invalid value` on the other hand lets them know that they have entered something which is not a valid input and encourage them correct it.

davidqc (approved):
OK if we can't reuse the ```cast next_url``` code in ```response action redirect/login```

davidqc:
image

PR #724, opened 2022-06-20
S3 save update

davidqc:
Rather than ```s3 get``` using a 2-step verification of file md5 (vulnerable to race conditions), move the logic into ```_s3_save``` where we use the ```http_save``` response headers to check md5 value in a single step.
Warn if we cannot check md5.
http_save tests would be better with mock responses but using httpbin as an interim.

```


Summary of Test Results
	Total	1823	Passed	1810	Skipped	13	Failed	0

Sourced 76 Test Files.


```


davidqc:
Updated to preserve existing return value of http_save
```


Summary of Test Results
	Total	1823	Passed	1810	Skipped	13	Failed	0

Sourced 76 Test Files.


```

adam-bethell (approved):

PR #731, opened 2023-01-13
Lock File Chunk Upload per User

nicky-johnstone:
Testing
---
- [x] Implemented on local project and uploaded ~10MB files in batches of 6. Repeated many times (40+) and the timing issue did not occur once.

### DIFF
--- a/tcl/file.tcl
+++ b/tcl/file.tcl
@@ -55,41 +55,51 @@ proc qc::file_upload {name chunk chunks file} {
 
     # Move the AOLserver generated tmp file to one we will keep
     file rename [ns_getformfile file] $tmp_file
-    if { [nsv_exists uploads $user_id] } {
-	set dict [nsv_get uploads $user_id]
-    } else {
-	set dict {}
-    }
-    dict set dict $id $chunk $tmp_file
-    dict set dict $id chunks $chunks
-    dict set dict $id filename $file
 
-    set complete true
-    foreach chunk [.. 0 $chunks-1] {
-	if { ![dict exists $dict $id $chunk] } {
-	    set complete false
-	    break
-	} else {
-	    lappend files /tmp/$id.$chunk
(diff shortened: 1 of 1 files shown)


tcl/file.tcl
```diff
@@ -55,41 +55,51 @@ proc qc::file_upload {name chunk chunks file} {
...
-    if { $complete } {
-	# Join parts together
-        ::try {
-            exec_proxy cat {*}$files > /tmp/$id
-        } finally {
-            # Clean up
-            foreach file $files {
-                file delete $file
+
+    ns_mutex eval [nsv_get locks $user_id] {
```
nicky-johnstone, line 63: https://naviserver.sourceforge.io/n/naviserver/files/ns_mutex.html#section3

tcl/file.tcl
```diff
@@ -55,41 +55,51 @@ proc qc::file_upload {name chunk chunks file} {
...
-    foreach chunk [.. 0 $chunks-1] {
-	if { ![dict exists $dict $id $chunk] } {
-	    set complete false
-	    break
-	} else {
-	    lappend files /tmp/$id.$chunk
-	}
+
+    if { ![nsv_exists locks $user_id] } {
+        nsv_set locks $user_id [ns_mutex create]
```
nicky-johnstone, line 60: We create a lock for each user and use an NSV array to track them.

It perhaps would be better to go a layer further and have a lock per file per user but as far as I'm aware there's no guarantee that the file ID generated by the caller is unique. It would also required storing a dict as the value, checking lock existence in the dict, and updating the NSV with a new dict. Naviserver 4.99.21+ supports [nsv_dict](https://naviserver.sourceforge.io/n/naviserver/files/nsv.html#8) which would be more suited for this purpose.

A lock per user has been sufficient in my testing so far.

PR #734, opened 2023-02-21
Add max keys to lsbucket

davidqc:
Allow ```qc::s3 lsbucket``` to be called with a ```-max_keys``` argument to limit the number of results returned.
Default when unspecified is 1000 records.

#### Tests

```
Tests ended at Tue Feb 21 14:45:03 GMT 2023
all.tcl:	Total	1844	Passed	1831	Skipped	13	Failed	0
Sourced 77 Test Files.
Number of tests skipped for each constraint:
	1	earlier_than_8.5
	12	knownBug


Summary of Test Results
	Total	1844	Passed	1831	Skipped	13	Failed	0

Sourced 77 Test Files.


```

### DIFF
--- a/tcl/s3.tcl
+++ b/tcl/s3.tcl
@@ -26,7 +26,10 @@ proc qc::s3 { args } {
         }
         ls {
             # usage: s3 ls 
-            set nodes [qc::s3_xml_select [qc::_s3_get "" ""] {/ns:ListAllMyBucketsResult/ns:Buckets/ns:Bucket}]
+            set nodes [qc::s3_xml_select \
+                        [qc::_s3_get "" ""] \
+                        {/ns:ListAllMyBucketsResult/ns:Buckets/ns:Bucket} \
+                      ]
             return [qc::lapply qc::s3_xml_node2dict $nodes]
         }
         lsbucket {
--- a/tcl/s3.tcl
+++ b/tcl/s3.tcl
@@ -30,16 +30,28 @@ proc qc::s3 { args } {
             return [qc::lapply qc::s3_xml_node2dict $nodes]
         }
         lsbucket {
-            # usage: s3 lsbucket bucket {prefix}
+            # usage: s3 lsbucket -max_keys ? bucket {prefix}
             # s3 lsbucket myBucket Photos/
-            if { [llength $args] == 1 || [llength $args] > 3 } {
-                error "Missing argument. Usage: qc::s3 lsbucket mybucket {prefix}"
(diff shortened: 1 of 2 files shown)


tcl/s3.tcl
```diff
@@ -26,21 +26,39 @@ proc qc::s3 { args } {
...
-            set nodes [qc::s3_xml_select [qc::_s3_get "" ""] {/ns:ListAllMyBucketsResult/ns:Buckets/ns:Bucket}]
+            set nodes [qc::s3_xml_select \
+                        [qc::_s3_get "" ""] \
+                        {/ns:ListAllMyBucketsResult/ns:Buckets/ns:Bucket} \
+                      ]
             return [qc::lapply qc::s3_xml_node2dict $nodes]
         }
         lsbucket {
-            # usage: s3 lsbucket bucket {prefix}
+            # usage: s3 lsbucket -max_keys ? bucket {prefix}
```
daniel-qcode, line 36: suggestions would be:
   1. Refactor qc::s3 to use TCL ensembles to provide the subcommands as `proc` definitions ... would avoid things like `set args [lrange $args 1 end]` when trying to pass along args
   2. Avoid using `qc::checks` which returns USER errors in APIs
   3. Think you said default max_keys was 1000 ... is that an upper limit that is enforced if the developer tried setting `-max_keys 2000` should we error or will S3 report error (or just return 1000)
   4. Helper proc to construct `object_string` (is this similar to URL query string, does it need URL encoding of parameters?)
   5. Maybe helper proc to process response

1. might be hard to do unless we
 * renamed existing legacy commands to make way for the new ensemble .. then initially just called those legacy procs from the new ensemble (legacy commands could be scheduled for further refactoring/deprecation over time).
 * agree on a new ensemble name for s3 commands

davidqc: Agree this whole proc need refactoring.

The API Docs
https://docs.aws.amazon.com/AmazonS3/latest/API/API_ListObjects.html
```
GET /?delimiter=Delimiter&encoding-type=EncodingType&marker=Marker&max-keys=MaxKeys&prefix=Prefix HTTP/1.1
Host: Bucket.s3.amazonaws.com
x-amz-request-payer: RequestPayer
x-amz-expected-bucket-owner: ExpectedBucketOwner
```

```qc::url``` would do the job but the ```s3_endpoint``` proc makes an assumption that a ```/``` should be appended so further refactoring needed.

daniel-qcode:
@davidqc can you close if nothing else needed in this PR, think all covered by the new PR that was just merged?

PR #757, opened 2023-08-29
Implement basic caching on IMDSv2 token

davidqc:
Board Ticket #
--------------
https://github.com/qcode-software/tlc/issues/7870

Git Release Type ( MAJOR | MINOR | PATCH )
--------------
(MINOR)

Description
--------------
Store IMDSv2 token in an environment variable until expiry.
Refresh token on expiry.

Test Results
--------------
```


Summary of Test Results
	Total	1887	Passed	1874	Skipped	13	Failed	0

Sourced 83 Test Files.


```


tcl/aws_metadata.tcl
```diff
@@ -0,0 +1,60 @@
+namespace eval qc {
+    namespace export \
+        aws_metadata \
+        aws_metadata_get \
```
daniel-qcode, line 4: do `aws_metadata_get` and `aws_metadata_token_refresh` need to be exported?

Only reason to export is if we want caller to be able to call using `aws_metadata_token_refresh` instead of fully qualified `qc::aws_metadata_token_refresh`.

davidqc: Updated

daniel-qcode (approved):

